PR #56, opened 2021-11-30
TFKUBE-51: Fixed the link to supported boundaries

nghazali:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,7 +1,7 @@
 blank_issues_enabled: false
```
nghazali, line 1: `.github/ISSUE_TEMPLATE/config.yml` and `.github/ISSUE_TEMPLATE/FEATURE.yml` was added before but because repo was not public was not visible. Now this feature is active and I just adjusted the link to the supported boundaries.

nanux (approved):

PR #65, opened 2021-12-03
TFKUBE-154: Add log into uninstall and makes log unique per install/uninstall run

nghazali:


pkg/scripts/uninstall.sh
```diff
@@ -121,7 +122,7 @@ destroy_tfstate() {
     if [ ${S3_BUCKET_EXISTS} -eq 0 ]
     then
       set +e
-      terraform destroy -auto-approve "${OVERRIDE_CONFIG_FILE}"
+      terraform destroy -auto-approve "${OVERRIDE_CONFIG_FILE}" | tee -a "${LOG_FILE}"
```
nanux, line 125: If you `tee` the log file, then the `$?` is the result of the `tee` command, which is usually `0` and not dependent on the `terraform destroy` result. If you want to check the outcome in the subsequent condition, you need to add `set -o pipefail` to the top of the script.

pkg/scripts/install.sh
```diff
@@ -7,8 +7,8 @@
 set -e
 CURRENT_PATH="$(pwd)"
 SCRIPT_PATH="$(dirname "$0")"
-LOG_FILE="${SCRIPT_PATH}/../../terraform-dc-install.log"
-LOG_TAGGING="${SCRIPT_PATH}/../../terraform-dc-asg-tagging.log"
+LOG_FILE="${SCRIPT_PATH}/../../terraform-dc-install_$(date '+%Y-%m-%d_%H-%M-%S').log"
```
nanux, line 10: What is the example of the final format?

pkg/scripts/uninstall.sh
```diff
@@ -87,7 +88,7 @@ destroy_infrastructure() {
   cd "${SCRIPT_PATH}/../../"
   set +e
   # Start destroying the infrastructure
-  terraform destroy "${OVERRIDE_CONFIG_FILE}"
+  terraform destroy "${OVERRIDE_CONFIG_FILE}" | tee "${LOG_FILE}"
```
nanux, line 91: The same as https://github.com/atlassian-labs/data-center-terraform/pull/65/files#r762721199

nanux (changes requested):

jjeongatl:
can you update .gitignore as well?

pkg/scripts/install.sh
nghazali: This is gonna be the sample of output files:

```
terraform-dc-install_2021-12-03_18-34-05.log

terraform-dc-uninstall_2021-12-03_18-17-56.log
```
For each time install or uninstall script runs we will have a separated log file, with date and time of running the script.

nanux (approved):

nanux:
> can you update .gitignore as well?

Good point, let's do it in the next PR.

PR #69, opened 2021-12-07
TFKUBE-139: Fixed the environment name variable validation and improved install/uninstall and logging

nghazali:


pkg/products/bamboo/variables.tf
```diff
@@ -4,11 +4,11 @@ variable "region_name" {
 }
 
 variable "environment_name" {
-  description = "Name of the cluster."
+  description = "Name of the environment."
   type        = string
   validation {
-    condition     = can(regex("^([a-zA-Z])+(([a-zA-Z]|[0-9])*-?)*$", var.environment_name))
-    error_message = "Invalid cluster name."
+    condition     = can(regex("^[a-zA-Z][a-zA-Z0-9\\-]{1,245}$", var.environment_name))
```
yzha645, line 10: Should be {1,24}?

nghazali: nice catch ;)

pkg/scripts/generate-variables.sh
```diff
@@ -1,67 +1,93 @@
...
+  echo ${CONFIG_FILE}
+  # extract S3 bucket, dynamodb, tags, and region from locals.tf
+  ENVIRONMENT_NAME=$(grep 'environment_name' ${CONFIG_FILE} | sed -nE 's/^.*"(.*)".*$/\1/p')
+  REGION=$(grep 'region' ${CONFIG_FILE} | sed -nE 's/^.*"(.*)".*$/\1/p')
+
+  # Get the AWS account ID
+  AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+
+  # Generates the unique s3 bucket and key names for the deployment for keeping the terraform state
+  S3_BUCKET="atlas-${ENVIRONMENT_NAME}-${REGION}-${AWS_ACCOUNT_ID}-tfstate"
```
yzha645, line 30: Are we creating s3 bucket per each environment? I vaguely remember that we create one bucket, and put different environment deployments into sub-folders, i.e. bucket key.

nghazali: That was for ages ago, as far as we handle multiple products in one environment now, we have decided to have a separated S3 for each environment to be more manageable and we can cleanup all resources including s3 independently.

yzha645: I see. yup, make sense. Thanks for the clarification! 

In this case, if user used all 25 characters as environment name, S3 bucket name will be trimmed to 64 chars in the following line. It then looks like `atlas-environmentname-ap-southeast-1-123456789012-tfst`. The last bit `tfst` looks a bit ugly to me... Since it is only 3-character difference, is it possible to limit the environment name to 22 chars long?

nghazali: I totally understand your point. However no one look at this names in normal circumstances. We can even drop the postfix at all. This case is the same for cluster name. We may drop part or all part of cluster name postfix which is totally okay. The most important thing is we still make the cluster name and s3 bucket name unique. For that reason I will leave the length of environment name to the maximum available length that keep the s3 and cluster names unique, in our case 25 characters. This gives the maximum flexibility to user to choose environment name.

yzha645: Where does 25 characters come from when you first decide it? cos 3-character difference doesn't seem significant to me. User wont lose too much flexibility. 
 
And you're absolutely right when looking terraform infra solely, probably resource name isn't that important. However, I'm more thinking from a user perspective. I wouldn't be surprised that user will have other workloads in the same aws account. Keeping the name consistent will provide user some pattern to easily opt in or filter out Atlassian deployment artifacts, maybe used for some reporting purpose or scheduled jobs to scan certain resources. `xxx-tfstate`, `xxx-tfstat` or `xxx.tfst` doesn't provide such consistency/pattern. And same apply for other aws resources we created. 

Another option I can think of is using `atl` instead of `atlas` as prefix, in this case, there are still 24 chars avail to user.

nghazali: Good point. What I did, considered all places we use environment name and picked the one with most constrains to define the maximum length for this item. 
Saying that, it comes from cluster naming - `atlassian-dc-<env>-cluster`.
As cluster name can be max 38 characters so considering sacrifice the postfix `-cluster`, we can have maximum 25 characters for environment name.

nghazali: I edited postfix and prefix for VPC, cluster, S3, and EFS names to make them consistent and also avoid postfix sacrifice and at the same time keep 25 length for environment name.

pkg/scripts/generate-variables.sh
```diff
@@ -27,7 +27,7 @@ set_variables() {
   AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
 
   # Generates the unique s3 bucket and key names for the deployment for keeping the terraform state
-  S3_BUCKET="atlas-${ENVIRONMENT_NAME}-${REGION}-${AWS_ACCOUNT_ID}-tfstate"
+  S3_BUCKET="atlas-${ENVIRONMENT_NAME}-${REGION}-${AWS_ACCOUNT_ID}-tfst"
```
yzha645, line 30:
```suggestion
  S3_BUCKET="atlas-tf-${ENVIRONMENT_NAME}-${REGION}-${AWS_ACCOUNT_ID}"
```

pkg/scripts/generate-variables.sh
```diff
@@ -1,67 +1,107 @@
...
+  # extract S3 bucket, dynamodb, tags, and region from locals.tf
+  ENVIRONMENT_NAME=$(grep 'environment_name' ${CONFIG_FILE} | sed -nE 's/^.*"(.*)".*$/\1/p')
+  REGION=$(grep 'region' ${CONFIG_FILE} | sed -nE 's/^.*"(.*)".*$/\1/p')
+
+  # Get the AWS account ID
+  AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+
+  # Generates the unique s3 bucket and key names for the deployment for keeping the terraform state
+  S3_BUCKET="atlas-${ENVIRONMENT_NAME}-${REGION}-${AWS_ACCOUNT_ID}-tfst"
+  BUCKET_KEY="${ENVIRONMENT_NAME}-${AWS_ACCOUNT_ID}"
```
yzha645, line 31: nit: Account id may not be necessary here.

pkg/scripts/generate-variables.sh
```diff
@@ -1,67 +1,107 @@
...
+
+  # length of the bucket name should be less than 64 characters
+  S3_BUCKET="${S3_BUCKET:0:63}"
```
yzha645, line 34: nit: may safely be removed.

yzha645 (approved):

pkg/scripts/generate-variables.sh
nghazali: To be consistent to other naming conventions I leave it as postfix

pkg/scripts/generate-variables.sh
nghazali: Right, but still is a safety factor. I didn't remove the same logic for cluster as well.

pkg/scripts/generate-variables.sh
nghazali: Yes, as this key is the only one in this bucket then environment name is sufficient.

PR #71, opened 2021-12-07
TFKUBE-170: Fix the default config file

nanux:
When the config file wasn't provided, the default wasn't set correctly and the `terraform apply` was missing variables.

Encountered by @ccorredor-atlassian

### DIFF
--- a/pkg/scripts/install.sh
+++ b/pkg/scripts/install.sh
@@ -6,7 +6,7 @@
 # -h : provides help to how executing this script.
 set -e
 set -o pipefail
-CURRENT_PATH="$(pwd)"
+ROOT_PATH="$(pwd)"
 SCRIPT_PATH="$(dirname "$0")"
 LOG_FILE="${SCRIPT_PATH}/../../terraform-dc-install_$(date '+%Y-%m-%d_%H-%M-%S').log"
 LOG_TAGGING="${SCRIPT_PATH}/../../terraform-dc-asg-tagging_$(date '+%Y-%m-%d_%H-%M-%S').log"
@@ -53,14 +53,14 @@ EOF
 process_arguments() {
   # set the default value for config file if is not provided
   if [ -z "${CONFIG_FILE}" ]; then
-    CONFIG_FILE="${SCRIPT_PATH}/../../config.tfvars"
+    CONFIG_FILE="config.tfvars"
   else
     if [[ ! -f "${CONFIG_FILE}" ]]; then
       echo "Terraform configuration file '${CONFIG_FILE}' is not found!"
       show_help
     fi
-    OVERRIDE_CONFIG_FILE="-var-file=${CONFIG_FILE}"
   fi
+  OVERRIDE_CONFIG_FILE="-var-file=${CONFIG_FILE}"
   echo "Terraform uses '${CONFIG_FILE}' to install the infrastructure."
 
   if [ ! -z "${UNKNOWN_ARGS}" ]; then
@@ -106,8 +106,6 @@ cleanup_terraform_backend_variables() {
 
 # Generates ./terraform-backend.tf and ./pkg/tfstate/tfstate-local.tf using the content of local.tf and current aws account
 generate_terraform_backend_variables() {
-  ROOT_FOLDER="${SCRIPT_PATH}/../.."
-
   echo "${ENVIRONMENT_NAME}' infrastructure deployment is started using ${CONFIG_FILE}."
 
   echo "Terraform state backend/variable files are missing."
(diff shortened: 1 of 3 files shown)


pkg/scripts/install.sh
```diff
@@ -53,14 +53,14 @@ EOF
 process_arguments() {
   # set the default value for config file if is not provided
   if [ -z "${CONFIG_FILE}" ]; then
-    CONFIG_FILE="${SCRIPT_PATH}/../../config.tfvars"
```
nghazali, line 56: All reference to files are based on the location of scripts - This is because of prevent to obligate user from running  the scripts from the root of repo. We better to keep this to be consistent with this rule.

pkg/scripts/install.sh
```diff
@@ -106,8 +106,6 @@ cleanup_terraform_backend_variables() {
 
 # Generates ./terraform-backend.tf and ./pkg/tfstate/tfstate-local.tf using the content of local.tf and current aws account
 generate_terraform_backend_variables() {
-  ROOT_FOLDER="${SCRIPT_PATH}/../.."
```
nghazali, line 109: The same as above

pkg/scripts/install.sh
```diff
@@ -120,13 +118,13 @@ create_tfstate_resources() {
   echo "Checking the terraform state."
   cd "${SCRIPT_PATH}/../tfstate"
   set +e
-  aws s3api head-bucket --bucket "${S3_BUCKET}"
+  aws s3api head-bucket --bucket "${S3_BUCKET}" 2>/dev/null
```
nghazali, line 121: thanks

pkg/scripts/install.sh
```diff
@@ -120,13 +118,13 @@ create_tfstate_resources() {
...
   S3_BUCKET_EXISTS=$?
   set -e
   if [ ${S3_BUCKET_EXISTS} -eq 0 ]
   then
     echo "S3 bucket '${S3_BUCKET}' already exists."
-    cd "${CURRENT_PATH}"
+    cd "${ROOT_PATH}"
```
nghazali, line 127: again this is not needed. The intention is user can run the script from any folder, the scripts will run and after run returns to the same folder. I mean we should not change the current folder as result of running the script.

pkg/scripts/install.sh
```diff
@@ -135,7 +133,7 @@ create_tfstate_resources() {
     sleep 5s
 
     echo "Migrating the terraform state to S3 bucket..."
-    cd "${CURRENT_PATH}"
+    cd "${ROOT_PATH}"
```
nghazali, line 136: The same as above

pkg/scripts/uninstall.sh
```diff
@@ -8,7 +8,7 @@
 
 set -e
 set -o pipefail
-CURRENT_PATH="$(pwd)"
+ROOT_PATH="$(pwd)"
```
nghazali, line 11: This is no guarantee if the current folder is root folder

nghazali (changes requested):

pkg/scripts/install.sh
```diff
@@ -53,14 +53,14 @@ EOF
...
-    CONFIG_FILE="${SCRIPT_PATH}/../../config.tfvars"
+    CONFIG_FILE="config.tfvars"
   else
     if [[ ! -f "${CONFIG_FILE}" ]]; then
       echo "Terraform configuration file '${CONFIG_FILE}' is not found!"
       show_help
     fi
-    OVERRIDE_CONFIG_FILE="-var-file=${CONFIG_FILE}"
   fi
+  OVERRIDE_CONFIG_FILE="-var-file=${CONFIG_FILE}"
```
nghazali, line 63: No need to move this out of if block. If there is no other config file is passed then we don't have to set `OVERRIDE_CONFIG_FILE` and is good to leave this blank. In this case it uses default config file which is `config.tfvars` from root folder of repo.

nanux: It won't as Carlos experienced firsthand. Because we have removed `.auto` from the config file name, the  config won't be picked up.

pkg/scripts/install.sh
nanux: I think users should run the scripts from the root, I would almost argue the `install.sh` and `uninstall.sh` scripts should be located in the root folder.

Yes, we should make it more flexible so you can call it from anywhere, but that doesn't work even in the current implementation.

pkg/scripts/install.sh
nanux: The `cd` was there, I just did the rename from `CURRENT_PATH` which is not self-explanatory to `ROOT_PATH` which refers to the root folder.

nanux:
I think this PR is now outdated after changes in #69 but I believe the problem with the default config file is still present. Since we don't use the `.auto` part in the config file, Terraform won't pick it up and end up running without any config file.

I also think moving the scripts (install and uninstall) into root would make them easier to discover and work with.

pkg/scripts/install.sh
nghazali: The bug has fixed in PR#69 and this feature is working fine (running from anywhere). It means we can safely move install and uninstall scripts to root and it expected to work with no further required change

pkg/scripts/install.sh
nghazali: It is not true, I just test it again and the default `config.tfvars` is picked if the replacement config file is not provided.

pkg/scripts/install.sh
nghazali: New version of the install script has no `cd` command at all. I removed all.

pkg/scripts/install.sh
nanux: Awesome, if this works now after the changes, I can close this PR.

PR #72, opened 2021-12-07
TFKUBE-172: Removed the 'desired_capacity' from replica number of csi controller

nghazali:


pkg/products/common/main.tf
```diff
@@ -23,7 +23,7 @@ module "efs" {
   region_name                  = var.region_name
   vpc                          = module.vpc
   eks                          = module.eks
-  csi_controller_replica_count = var.desired_capacity
+  csi_controller_replica_count = 1
```
yzha645, line 26: Just a question, is 1 enough if the initial config file sets desired_capacity to say 10?

nghazali: I tested with two nodes and setting this replica count to 1 was working fine. But we may need more tests. What is obvious, if 'csi_controller_replica_count' changes then efs controller will stuck in updating stage and terraform fails to complete install.

yzha645: yea, since testing needs 50 agents, potentially we may set desired_capacity to 10 or more. I think we need to verify this, maybe in the performance test story. 

Another way I can think of is putting the first time value to a local variable, something like
```
if local.csi_controller_replica_count is Null {
    local.csi_controller_replica_count = var.desired_capacity
}
```
But it's just a thought, haven't tested yet.

nghazali: Anyhow I found we can't change `desired_capacity` at all. I mean we can change, but this change does not affected in number of nodes when applied in terraform. Scaling is matter of manual change in AWS console. So, anyway we may need to remove `desired_capacity` from configurable items in config file. 
For now we safe to change this to `1` and we will revisit this after performance test if is needed.

yzha645 (approved):

PR #76, opened 2021-12-10
TFKUBE-187: Fixed the unsupported attribute error in install/uninstall when the infrastructure is not set

nghazali:


pkg/modules/AWS/eks/locals.tf
```diff
@@ -0,0 +1,4 @@
+
+locals {
+  cluster_asg_name = try(module.eks.node_groups.appNodes.resources[0].autoscaling_groups[0].name, null)
```
nanux, line 3: Can we add a comment on why we have `try` here?

nghazali: done

PR #80, opened 2021-12-13
TFKUBE-183: Fully pre seed bamboo helm release

yzha645:
As before Bamboo 8.1, the current Terraform solution will break as it's partially pre-seeded: 
- database (+)
- local home (+)
- shared home (+)
- ingress (+)
- license (-)
- admin creds (-)

This PR fills the gap to mandate the configuration of license and sysAdminCredentials. 

Note: This PR doesn't allow the flexibility to skip/partially pre-seed Bamboo as there is some limitation from Bamboo side. We may enhance the solution to make these values optional in the future.


pkg/products/bamboo/secrets.tf
```diff
@@ -0,0 +1,43 @@
...
+
+  data = {
+    license = var.license
+  }
+}
+
+################################################################################
+# Kubernetes secret to store system admin credentials
+################################################################################
+resource "kubernetes_secret" "admin_secret" {
```
nghazali, line 32: Also this one should go to they bamboo namespace. Each product may have different admin info.

pkg/products/bamboo/variables.tf
```diff
@@ -50,4 +50,34 @@ variable "db_instance_class" {
...
   type        = number
+}
+
+variable "license" {
+  description = "Bamboo license."
+  type        = string
+  sensitive   = true
+}
+
+variable "admin_username" {
```
nghazali, line 61: Here the only sensitive items are license and admin password. All other variables are normal and could be seen as plain text.

variables.tf
```diff
@@ -68,4 +68,34 @@ variable "db_iops" {
...
+variable "bamboo_license" {
+  description = "Bamboo license."
+  type        = string
+  sensitive   = true
+}
+
+variable "bamboo_admin_username" {
+  description = "Bamboo system administrator username."
+  type        = string
+  sensitive   = true
```
nghazali, line 82: `bamboo_admin_username`, `bamboo_admin_display_name`, and `bamboo_admin_email_address` are not sensitive

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -157,3 +157,43 @@ db_allocated_storage = 100
 ```terraform
 db_iops = 1000
 ```
+
+### Bamboo License
+`bamboo_license` takes the license key of Bamboo product.
+
+To avoid storing license in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_`.
```
nanux, line 164:
```suggestion
To avoid storing license in a plain text file, we recommend storing it in an environment variable prefixed with [`TF_VAR_`](https://www.terraform.io/docs/cli/config/environment-variables.html#tf_var_name).
```

config.tfvars
```diff
@@ -37,4 +37,15 @@ db_instance_class    = "db.t3.micro"
 db_allocated_storage = 1000
 db_iops              = 1000
 
-
+# Bamboo license
+# To avoid storing license in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_` and keep the below line commented out
```
nanux, line 41:
```suggestion
# To avoid storing license in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_` (i.e. `TF_VAR_bamboo_license`) and keep the below line commented out
```

pkg/products/bamboo/locals.tf
```diff
@@ -27,4 +27,25 @@ locals {
...
+      license = {
+        secretName = kubernetes_secret.license_secret.metadata[0].name
+        secretKey  = "license"
+      }
+    }
+  })
+
+  admin_settings = yamlencode({
+    bamboo = {
+      sysadminCredentials = {
```
nghazali, line 42: We may not need to put all this items in the secret.

nanux:
What about tests?

nanux:
Can we also add some tests?

nghazali (changes requested):

pkg/products/bamboo/variables.tf
```diff
@@ -50,4 +50,34 @@ variable "db_instance_class" {
...
+variable "admin_password" {
+  description = "System administrator password."
+  type        = string
+  sensitive   = true
+}
+
+variable "admin_display_name" {
+  description = "System administrator display name."
+  type        = string
+  sensitive   = true
```
nanux, line 76: Why this is sensitive?

pkg/products/bamboo/variables.tf
```diff
@@ -50,4 +50,34 @@ variable "db_instance_class" {
...
+}
+
+variable "admin_email_address" {
+  description = "System administrator email address."
+  type        = string
+  sensitive   = true
```
nanux, line 82: Why this is sensitive?

pkg/products/bamboo/variables.tf
```diff
@@ -50,4 +50,34 @@ variable "db_instance_class" {
...
+variable "license" {
+  description = "Bamboo license."
+  type        = string
+  sensitive   = true
+}
+
+variable "admin_username" {
+  description = "System administrator username."
+  type        = string
+  sensitive   = true
```
nanux, line 64: Why this is sensitive?

variables.tf
nanux, line 82: The same questions as above - Why this is sensitive?

yzha645:
@nanux yup, working on expanding end to end tests right now.

pkg/products/bamboo/secrets.tf
```diff
@@ -0,0 +1,43 @@
...
+  data = {
+    username = module.database.rds_master_username
+    password = module.database.rds_master_password
+  }
+}
+
+################################################################################
+# Kubernetes secret to store license
+################################################################################
+resource "kubernetes_secret" "license_secret" {
```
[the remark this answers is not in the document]
yzha645: yup, make sense!

pkg/products/bamboo/locals.tf
nanux: The helm chart is expecting it being in the secret - https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bamboo/values.yaml#L398

config.tfvars
```diff
@@ -37,4 +37,15 @@ db_instance_class    = "db.t3.micro"
...
+# If deploy in a non-prod environment and plain-text value does not expose a security risk, feel free to uncomment the following line and supply the license here
```
nanux, line 42: I would skip the `non-prod` environment message as we don't recommend this for prod environments.

config.tfvars
```diff
@@ -37,4 +37,15 @@ db_instance_class    = "db.t3.micro"
...
+#bamboo_license = "<license key>"
+
+# Bamboo system admin credentials
+# To avoid storing system admin credentials in a plain text file, we recommend storing them in environment variables prefixed with `TF_VAR_` and keep the below 4 lines commented out
```
nanux, line 46: I think this is true only for the password, the rest are not too sensitive.

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -157,3 +157,43 @@ db_allocated_storage = 100
 ```terraform
 db_iops = 1000
 ```
+
+### Bamboo License
```
nanux, line 161: I think somewhere we should create a `Sensitive Data` section/warning and link https://www.terraform.io/docs/language/state/sensitive-data.html and https://kubernetes.io/docs/concepts/configuration/secret/#security-properties so users can read about implications.

I find `Kubernetes Secrets` the most misleading naming in the whole ecosystem.

pkg/products/bamboo/variables.tf
yzha645: yea, I'm not too sure about this part either. I'm basically open to either way. 

The reason I'm marking it as sensitive is because all of them are stored in Kubernetes secrets. And I guess whether it's sensitive or nor is not just about the nature of the value itself, but how we implement/store it, in this case, as a Kubernetes secret. 

@nanux @jjeongatl What's your opinion?

nanux (changes requested):

pkg/products/bamboo/locals.tf
yzha645: As from Bamboo Helm chart, it requires all values stored in the secret. 
https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bamboo/values.yaml#L398

pkg/products/bamboo/variables.tf
nanux: I would make just password sensitive.

yzha645: Cool~ insensitive it is then.

nghazali (approved):

nanux (approved):

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
...
+  fi
+
+  if [ -n "${POPULATED_ADMIN_PWD}" ];  then
+    if [ -n "${TF_VAR_bamboo_admin_password}" ]; then
+      echo "Please provide admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'."
+      HAS_VALIDATION_ERR=1
+    fi
+  fi
+
+  if [ ! -n "${HAS_VALIDATION_ERR}" ]; then
```
nanux, line 121: This didn't work for me, I think we should convert the `HAS_VALIDATION_ERR` variable to a boolean type and then compare it with:

```
if [ "${HAS_VALIDATION_ERR}" == true ]; then
```

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
 verify_configuration_file() {
   log "Verifying the config file."
 
+  HAS_VALIDATION_ERR=
```
nanux, line 80: This can be boolean
```suggestion
  HAS_VALIDATION_ERR=false
```

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
...
 
+  # check license and admin password
+  export POPULATED_LICENSE=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_license')
+  export POPULATED_ADMIN_PWD=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_admin_password')
+
   if [ "${#ENVIRONMENT_NAME}" -gt 25 ]; then
     log "The environment name '${ENVIRONMENT_NAME}' is too long(${#ENVIRONMENT_NAME} characters)."
     log "Please make sure your environment name is less than 25 characters"
-    exit 1
+    HAS_VALIDATION_ERR=1
```
nanux, line 94:
```suggestion
    HAS_VALIDATION_ERR=true
```

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
...
+
+  if [ -n "${POPULATED_ADMIN_PWD}" ];  then
+    if [ -n "${TF_VAR_bamboo_admin_password}" ]; then
+      echo "Please provide admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'."
+      HAS_VALIDATION_ERR=1
+    fi
+  fi
+
+  if [ ! -n "${HAS_VALIDATION_ERR}" ]; then
+    exit 1
```
nanux, line 122: I would also `log "There was a problem with the configuration file. Aborting execution" "ERROR"` line above.

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
...
     log "Then re-run the install.sh to deploy the infrastructure."
     echo
     log "${INVALID_CONTENT}"
-    exit 0
+    HAS_VALIDATION_ERR=1
+  fi
+
+  if [ -n "${POPULATED_LICENSE}" ];  then
+    if [ -n "${TF_VAR_bamboo_license}" ]; then
+      echo "Please provide license in config file, or export it to the environment variable 'TF_VAR_bamboo_license'."
```
nanux, line 109:
```suggestion
      log "Please provide license in config file, or export it to the environment variable 'TF_VAR_bamboo_license'."
```

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
...
+      HAS_VALIDATION_ERR=1
+    fi
+  fi
+
+  if [ -n "${POPULATED_ADMIN_PWD}" ];  then
+    if [ -n "${TF_VAR_bamboo_admin_password}" ]; then
+      echo "Please provide admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'."
```
nanux, line 116:
```suggestion
      log "Please provide admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'."
```

install.sh
nghazali: I don't think this message adds much as we get to this condition only if we already caught some invalid condition earlier, and for each invalid condition we already had logged the proper message before. Here the only action is aborting the process.

nanux: Previously the whole execution printed:

```
2021-12-20 20:26:12 [INFO] Terraform will use '/Users/abrokes/code/data-center-terraform/abrokes-config.tfvars' to install the infrastructure.
2021-12-20 20:26:12 [INFO] Verifying the config file.
```
I know it was due to a bad condition but mistakes will happen in scripting and it is good to give a good guide where the execution was interrupted. And printing it out as `ERROR` is also useful.

nghazali: Actually in case of any condition we log the specific case before turn the error flag to `true`. 
such as:
```
2021-12-20 20:26:12 [INFO] Terraform will use '/Users/abrokes/code/data-center-terraform/abrokes-config.tfvars' to install the infrastructure.
2021-12-20 20:26:12 [ERROR] Configuration file 'abrokes-config.tfvars' is not valid.
2021-12-20 20:26:12 [INFO] Please make sure your environment name is less than 25 characters.
```

And if there are more than one issue, all will list before exit.

Again, wrapping up all error condition with a final error message is not bad idea, but if we don't have that message still we won't miss much.

install.sh
```diff
@@ -77,26 +77,50 @@ process_arguments() {
 verify_configuration_file() {
   log "Verifying the config file."
 
+  HAS_VALIDATION_ERR=false
```
nghazali, line 80:
```suggestion
  HAS_VALIDATION_ERR=
```

install.sh
```diff
@@ -77,26 +77,50 @@ process_arguments() {
...
 
+  # check license and admin password
+  export POPULATED_LICENSE=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_license')
+  export POPULATED_ADMIN_PWD=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_admin_password')
+
   if [ "${#ENVIRONMENT_NAME}" -gt 25 ]; then
     log "The environment name '${ENVIRONMENT_NAME}' is too long(${#ENVIRONMENT_NAME} characters)."
     log "Please make sure your environment name is less than 25 characters"
-    exit 1
+    HAS_VALIDATION_ERR=true
```
nghazali, line 94:
```suggestion
    HAS_VALIDATION_ERR=1
```

install.sh
```diff
@@ -77,26 +77,50 @@ process_arguments() {
...
-    log "Configuration file '${CONFIG_ABS_PATH}' is not valid."
+    log "Configuration file '${CONFIG_ABS_PATH##*/}' is not valid."
     log "Terraform uses this file to generate customised infrastructure for '${ENVIRONMENT_NAME}' on your AWS account."
-    log "Please modify '${CONFIG_ABS_PATH}' using a text editor and complete the configuration. "
+    log "Please modify '${CONFIG_ABS_PATH##*/}' using a text editor and complete the configuration. "
     log "Then re-run the install.sh to deploy the infrastructure."
     echo
     log "${INVALID_CONTENT}"
-    exit 0
+    HAS_VALIDATION_ERR=true
```
nghazali, line 104:
```suggestion
    HAS_VALIDATION_ERR=1
```

install.sh
```diff
@@ -77,26 +77,50 @@ process_arguments() {
...
+  fi
+
+  if [ -n "${POPULATED_LICENSE}" ];  then
+    if [ -n "${TF_VAR_bamboo_license}" ]; then
+      log "Please provide license in config file, or export it to the environment variable 'TF_VAR_bamboo_license'."
+      HAS_VALIDATION_ERR=true
```
nghazali, line 110:
```suggestion
      HAS_VALIDATION_ERR=1
```

install.sh
```diff
@@ -77,26 +77,50 @@ process_arguments() {
...
+    fi
+  fi
+
+  if [ -n "${POPULATED_ADMIN_PWD}" ];  then
+    if [ -n "${TF_VAR_bamboo_admin_password}" ]; then
+      log "Please provide admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'."
+      HAS_VALIDATION_ERR=true
```
nghazali, line 117:
```suggestion
      HAS_VALIDATION_ERR=1
```

install.sh
```diff
@@ -77,26 +77,50 @@ process_arguments() {
...
+    fi
+  fi
+
+  if [ "${HAS_VALIDATION_ERR}" == true ]; then
```
nghazali, line 121:
```suggestion
  if [ -n "${HAS_VALIDATION_ERR}" ]; then
```

nghazali (changes requested):

install.sh
nanux: I am sorry, that is my mistake. It should be:

```
HAS_VALIDATION_ERR="false"
```
or we can have 
```
HAS_VALIDATION_ERR="yes"
```

I think the scripts are hard to read and using the `1` doesn't really help. There are other options than using numbers. https://stackoverflow.com/a/47092826

Sorry @yzha645 for leading you astray.

nghazali: Using `1` and `0` for flags is used widely in the scripts so using this will keep consistency. 
We can refactor this later when we will add other products if is needed.

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
 verify_configuration_file() {
   log "Verifying the config file."
 
+  HAS_VALIDATION_ERR=false
```
nanux, line 80: @yzha645 you need to make this `HAS_VALIDATION_ERR=` as discussed over Slack, let's use the `1` approach.

yzha645: Good spot. Thank you! Fixed.

install.sh
```diff
@@ -77,26 +77,49 @@ process_arguments() {
...
+  export POPULATED_LICENSE=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_license')
+  export POPULATED_ADMIN_PWD=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_admin_password')
+
   if [ "${#ENVIRONMENT_NAME}" -gt 25 ]; then
-    log "The environment name '${ENVIRONMENT_NAME}' is too long(${#ENVIRONMENT_NAME} characters)."
-    log "Please make sure your environment name is less than 25 characters"
-    exit 1
+    log "The environment name '${ENVIRONMENT_NAME}' is too long(${#ENVIRONMENT_NAME} characters)." "ERROR"
+    log "Please make sure your environment name is less than 25 characters." 
+    HAS_VALIDATION_ERR=
```
nghazali, line 94:
```suggestion
    HAS_VALIDATION_ERR=1
```

nghazali (changes requested):

nghazali (approved):

PR #81, opened 2021-12-13
TFKUBE-195- Fixed the root folder in generate-variables.sh and cleanup.sh and install.sh

nghazali:


nanux:
Do we ever run the scripts on their own? I think the amount of code we are adding to handle the use cases when it is executed from root or from the script folder is getting non-negligible and mainly complicate reasoning about the scripts themselves. Should we just merge the cleanup into `uninstall.sh` and figure out how to share `generate-variables.sh` between scripts? I think with more code we put into the scripts the more I am concerned that they will get brittle and cause a bad user experience.

nghazali:
Cleanup could be used directly as a part of troubleshooting tool. This script will be redundant if we include in installer and uninstaller (original implementation). 

This change basically is just adding a path to root folder which was pruned before. I was trimmed some code to make it cleaner but I found `$0` in bash points to the folder of the first script that is executed in nested calls. So I have to revert this to make sure this scripts will work both as nested or standalone scripts.

nghazali:
Just I want to point you to this fact that this repo should be handle multiple environment, and it means we need to cleanup and generate variable files in both install and uninstall as we may have a different config file since last run.

nanux (approved):

nanux:
I am not fully comfortable with the implementation but we need to get it fixed as the scripts currently don't work locally so I am approving. I think we should make a group review of the scripts to spread the knowledge and make them less error-prone.

PR #82, opened 2021-12-13
TFKUBE-158 Add more content for troubleshooting page

nghazali:


docs/docs/troubleshooting/TROUBLESHOOTING.md
```diff
@@ -2,75 +2,93 @@
...
 
-## Interrupted uninstallation
+## Uninstall - Invalid configuration file   
+  
+**Symptom**
+If you try to uninstall an environment by using a different configuration file than the one you used to install it or by using a different version of the code, you may encounter some issues during uninstallation. In most cases, Terraform reports that the resource cannot be removed because it's in use.
 
-### Symptom
+**Solution**
+Idenity the resource and delete it manually from the AWS console, and then restart the uninstallation process.
```
errcode1202, line 35: Just wondering if it would be nicer to use an expand macro here? Something like...

```

???tip "Uninstall - Invalid configuration file "

**Symptom**
If you try to uninstall an environment by using a different configuration file than the one you used to install it or by using a different version of the code, you may encounter some issues during uninstallation. In most cases, Terraform reports that the resource cannot be removed because it's in use.

**Solution**
Idenity the resource and delete it manually from the AWS console, and then restart the uninstallation process.


```

docs/docs/troubleshooting/TROUBLESHOOTING.md
```diff
@@ -2,75 +2,93 @@
...
-### Solution
+Error: context deadline exceeded
 
-Idenity the resource and delete it manually from the AWS console, and then restart the uninstallation process. 
+Error: Persistent volume claim atlassian-dc-bamboo-share-home-pvc still exists with 
+```
+**Solution**
+If a bamboo pod termination stalls, it will block pvc and pv deletion. 
+To fix this problem we need to terminate product pod first and run uninstall command again.
+```
```
errcode1202, line 53: indicate the code type here using `shell` i.e.

```

  ```shell
  kubectl delete pod <bamboo-pod> -n bamboo --force
  ```

File: docs/docs/troubleshooting/TROUBLESHOOTING.md [hunk 3]
```diff
@@ -2,75 +2,93 @@
...
+kubectl delete pod <bamboo-pod> -n bamboo --force
+```
+To see the stalled bamboo pod name you can run the following command:
```
errcode1202, line 56: Capitaliize names, `Bamboo`

docs/docs/troubleshooting/TROUBLESHOOTING.md
```diff
@@ -2,75 +2,93 @@
...
+```
+kubectl get pods -n bamboo 
+```
 
-## Issues while accessing AWS
+## Uninstall: Suspended ASG
+If for any reason Auto Scaling Group gets suspended, AWS does not allow terraform to delete the node group.
```
errcode1202, line 62: Capitaliize names, `Terraform`

docs/docs/troubleshooting/TROUBLESHOOTING.md
```diff
@@ -2,75 +2,93 @@
...
+```
+kubectl get pods -n bamboo 
+```
 
-## Issues while accessing AWS
+## Uninstall: Suspended ASG
+If for any reason Auto Scaling Group gets suspended, AWS does not allow terraform to delete the node group. 
+
+**Symptom**
+Uninstall process get interrupted with the following error:
```
errcode1202, line 65: `get` - `get's`

docs/docs/troubleshooting/TROUBLESHOOTING.md
```diff
@@ -80,33 +98,25 @@ Acquiring state lock. This may take a few moments...
...
-
+**Solution**
 Forcibly unlock the state by running the following command:
 
 ```shell 
 terraform force-unlock <ID>
 ```
 
-Where `<ID>` is the value that appears in the error message.
+Where <ID> is the value that appears in the error message.
```
errcode1202, line 117: Add back ticks to `<ID>`

errcode1202 (changes requested):

docs/docs/troubleshooting/TROUBLESHOOTING.md
nghazali: That is a very nice suggestion indeed. The reason I have subjects as a title is to include them in table of content. In this case when the content of this page is long (which is) then they can easily jump into the issue by clicking on TOC in right side of the page.

errcode1202 (approved):

PR #91, opened 2021-12-20
TFKUBE-180: Import dataset

nanux:


variables.tf
```diff
@@ -54,20 +54,26 @@ variable "domain" {
...
   description = "The requested number of I/O operations per second that the DB instance can support."
-  type        = number
   default     = 1000
+  type        = number
+}
+
+variable "dataset_url" {
+  description = "URL of the dataset to restore in the Bamboo instance"
+  type        = string
+  default     = "https://centaurus-datasets.s3.amazonaws.com/bamboo/dcapt-bamboo.zip"
```
nghazali, line 76: I am thinking to make it optional. If we can find a way to default this to null or empty string and in case the user does not change it in initial configuration, then we can ignore the data import and otherwise  perform `dataset.tf` and import it in `helm_release.tf`.

pkg/products/bamboo/helm_release.tf
```diff
@@ -1,6 +1,7 @@
 # Install helm chart for Bamboo Data Center.
 
 resource "helm_release" "bamboo" {
+  depends_on = [kubernetes_job.import_dataset]
```
nghazali, line 4: How long does it take to import get completed with given dataset? Any timeout or issue has raised in your test?

pkg/products/bamboo/dataset.tf
```diff
@@ -0,0 +1,35 @@
...
+        restart_policy = "Never"
+        volume {
+          name = "shared-home"
+          persistent_volume_claim {
+            claim_name = kubernetes_persistent_volume_claim.atlassian-dc-bamboo-share-home-pvc.metadata[0].name
+          }
+        }
+      }
+    }
+    active_deadline_seconds = "600"
```
nghazali, line 31: What is active deadline? Is it different with import timeout? If this is the same is 10 min enough for import the complete given dataset including 20-30% buffer?

variables.tf
nanux: You are right, yesterday after testing it was clear we need to make it optional as it takes ~30 mins to restore this dataset. I have done the change.

pkg/products/bamboo/helm_release.tf
nanux: The import dataset job is ~10 seconds (downloading from s3 and saving the file in shared home). The actual import of the dataset takes around ~30 minutes and I am now changing the timeouts to make it possible. Currently, the `helm_release` will time out but I have increased the timeout but the testing is PITA as the whole cluster needs to be recreated.

pkg/products/bamboo/dataset.tf
nanux: This just for the job - it is actually too long probably, right now the dataset takes ~10 seconds but because this can be used for any dataset I will keep it 10mins just to be on the safe side.

yzha645 (approved):

pkg/products/bamboo/dataset.tf
```diff
@@ -0,0 +1,42 @@
+resource "kubernetes_job" "import_dataset" {
+  count = var.dataset_url != null ? 1 : 0
```
nghazali, line 2: 👍

nghazali (approved):

PR #92, opened 2021-12-21
TFKUBE-160: Bamboo agents

yzha645:
This PR adds Bamboo agent Helm release to bamboo module, so that agents pods can be deployed at the same time as bamboo instance. 
User is able to specify number of agents from configuration file.


docs/docs/userguide/CONFIGURATION.md
```diff
@@ -194,6 +194,16 @@ db_allocated_storage = 100
...
 ```
 
+### Number of Bamboo agents
+`number_of_bamboo_agents` sets the number of remote agents to be launched. 
+
+```terraform
+number_of_bamboo_agents = 50
+```
+
+* Note: This number should not be greater than the value of Edition in the issued license.
```
nanux, line 204: I think `Edition` is an internal concept, this should be something along the lines:

```suggestion
* Note: This number should not be greater than the number of allowed remote agents in your license.
```
Screen Shot 2021-12-21 at 4 15 29 pm

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -194,6 +194,16 @@ db_allocated_storage = 100
...
 
+### Number of Bamboo agents
+`number_of_bamboo_agents` sets the number of remote agents to be launched. 
+
+```terraform
+number_of_bamboo_agents = 50
+```
+
+* Note: This number should not be greater than the value of Edition in the issued license. 
+  Any exceeded number of agents won't be able to join.
```
nanux, line 205:
```suggestion
  Any agents exceeding the allowed number won't be able to join.
```

pkg/products/bamboo/helm_release.tf
```diff
@@ -59,4 +59,34 @@ data "kubernetes_service" "bamboo" {
...
+  namespace  = kubernetes_namespace.bamboo.metadata[0].name
+  repository = "https://atlassian.github.io/data-center-helm-charts"
+  chart      = "bamboo-agent"
+  version    = "0.0.2"
+
+  depends_on = [helm_release.bamboo]
+
+  values = [
+    yamlencode({
+      replicaCount = var.number_of_agents
```
nanux, line 75: How does this work if it is 0?

Is defining `number_of_agents = 0` a valid way how to disable agents?

variables.tf
```diff
@@ -95,4 +95,10 @@ variable "bamboo_admin_display_name" {
 variable "bamboo_admin_email_address" {
   description = "Bamboo system administrator email address."
   type        = string
+}
+
+variable "number_of_bamboo_agents" {
+  description = "Number of Bamboo remote agents."
```
nanux, line 101: Should we validate for the value being `>= 0`? (if we allow 0, see comment above).

pkg/products/bamboo/helm_release.tf
```diff
@@ -59,4 +59,34 @@ data "kubernetes_service" "bamboo" {
...
     namespace = kubernetes_namespace.bamboo.metadata[0].name
   }
+}
+
+resource "helm_release" "bamboo_agent" {
+  name       = "bamboo-agent"
+  namespace  = kubernetes_namespace.bamboo.metadata[0].name
+  repository = "https://atlassian.github.io/data-center-helm-charts"
+  chart      = "bamboo-agent"
+  version    = "0.0.2"
```
nanux, line 69: We should extract this into `locals` and use the same version for `bamboo` and `bamboo-agent` `helm_release` resources.

pkg/products/bamboo/helm_release.tf
yzha645: yes, `replicaCount = 0` will disable agents. i.e. the helm release will be installed, but no agent pod will be initiated. 

I was thinking about not install agent chart at all when this variable is set to 0. But had a rethink, installing Helm release with replicaCount = 0 will actually save user some time preparing agent's values.yaml file. To deploy agents later, user can just run kubectl command to scale up the deployments.

variables.tf
yzha645: Added.

pkg/products/bamboo/helm_release.tf
yzha645: Done.

pkg/products/bamboo/helm_release.tf
```diff
@@ -59,4 +59,34 @@ data "kubernetes_service" "bamboo" {
     name      = "bamboo"
     namespace = kubernetes_namespace.bamboo.metadata[0].name
   }
+}
+
+resource "helm_release" "bamboo_agent" {
+  name       = "bamboo-agent"
+  namespace  = kubernetes_namespace.bamboo.metadata[0].name
+  repository = "https://atlassian.github.io/data-center-helm-charts"
```
nghazali, line 67: Would you please create a local variable for atlassian helm chart repository and use the variable here and in `helm_release.bamboo`

pkg/products/bamboo/helm_release.tf
nghazali: It worth to point to this in documentation. (`CONFIGURATION.md`  and/or `config.tfvars`)?

pkg/products/bamboo/locals.tf
```diff
@@ -1,5 +1,6 @@
 locals {
-  product_name = "bamboo"
+  product_name       = "bamboo"
+  helm_chart_version = "0.0.2"
```
nghazali, line 3: I am thinking if we move the helm chart version as one of the configurable items. It makes the turn key solution independent of product version. We already have enough complexity in upgrade process. Mixing the infrastructure and product will make it even more complex. 
This is an idea and no action is required here. But we can open a discussion to consider the upgrade matter with terraform project.

variables.tf
```diff
@@ -95,4 +95,14 @@ variable "bamboo_admin_display_name" {
...
+  default     = 50
+  type        = number
+  validation {
+    condition     = var.number_of_bamboo_agents >= 0
+    error_message = "Number of agents must be greater than 0."
```
nghazali, line 106:
```suggestion
    error_message = "Number of agents must be greater or equal 0."
```

nghazali (changes requested):

nghazali (approved):

PR #103, opened 2022-01-17
TFKUBE-224: Added status test and implemented methods for pause/resume the server

nghazali:


test/e2etest/bamboo_test.go
```diff
@@ -289,3 +289,12 @@ func GetPageContent(t *testing.T, url string) []byte {
 	assert.NoError(t, err, "Error reading response body")
 	return content
 }
+
+//PostPageContent performs a post request the given content using the given url
+func PostPageContent(t *testing.T, url string, contentType string, body io.Reader) {
```
yzha645, line 294: 1. Can we change function name to something more precise? Post page content doesn't feel quite right to me. 
A good function name should be self-explanatory, and theoretically should not need any comment.  

2. If none of the caller uses `body`, maybe we can omit it here as an input parameter? 

3. Noticing the same for function `GetPageContent`, any reason we're making them public? Looks like they're only been used within the package.

test/e2etest/bamboo_test.go
```diff
@@ -289,3 +289,12 @@ func GetPageContent(t *testing.T, url string) []byte {
...
+	resp, err := http.Post(url, contentType, body)
+	require.NoError(t, err, "Error accessing url: %s", url)
+	defer resp.Body.Close()
+
+	assert.NoError(t, err, "Error reading response body")
```
yzha645, line 299: Since we already asserted `err` on line 296 using `require.NoError`, does it mean this line (line 299) will never be executed? Seems we are asserting the same err object twice.

test/e2etest/installer_test.go
```diff
@@ -25,12 +25,24 @@ func TestInstaller(t *testing.T) {
...
+	assertStatusEndpoint(t, testConfig, "RUNNING")
 
 	// Uninstall and cleanup the environment
 	runUninstallScript(testConfig.ConfigPath)
 }
 
+func assertStatusEndpoint(t *testing.T, testConfig TestConfig, expectedStatus string) {
+	statusUrl := "rest/api/latest/status"
+	url := fmt.Sprintf("https://%s.%s.%s/%s", product, testConfig.EnvironmentName, domain, statusUrl)
+	content := fmt.Sprintf("%s", GetPageContent(t, url))
```
yzha645, line 43: `fmt.Sprintf` is probably unnecessary here. Something like this might be sufficient: 
```
        content := GetPageContent(t, url)
	assert.Contains(t, string(content), "<title>All projects - Atlassian Bamboo</title>")
	assert.Contains(t, string(content), "totalRecords: 1")
```

yzha645 (changes requested):

test/e2etest/installer_test.go
nghazali: This is needed as `GetPageContent` returns `byte[]` and we need `string`. This actually converts `byte[]` to `string`.

test/e2etest/bamboo_test.go
nghazali: I followed existing GetPageContent and you are right. Those are not needed to be public and I renamed the method to `sendPostRequest`. 

We still need to keep the body as this is part of post request convention. We may not using them for our cases but good to have just in case if we want to use in future.

test/e2etest/installer_test.go
yzha645: Doesn't `string(content)` also convert `[]byte` to `string`? 
Just curious, from Google I found most answers use the above way I copied from our repo, i.e. `string(content)`. https://yourbasic.org/golang/convert-string-to-byte-slice/ 
Did you choose to use `fmt.Sprintf` because it's more efficient? I guess both are workable, and probably doesn't matter so much in our case given we're not too worried about performance. Just trying to close the gap here if you choose `fmt.Sprintf` for a reason.

nghazali: Generally in type conversion to string the best practise is using string builder. This is specially efficient when we creating the string accumulatively to avoid creating multiple short life objects. In our case we convert just once but still better use the best practice.

yzha645: Awesome. Good to know~ 🥇

yzha645 (approved):

PR #107, opened 2022-01-18
TFKUBE-48: Go linter

yzha645:
This PR added Golang linter to pre-commit hook and Github action for both unit test and e2e test. 
Will also fix linter errors for existing code.


.github/workflows/e2e-test.yaml
```diff
@@ -40,7 +40,15 @@ jobs:
         run: |
           go get -v -t -d ./... && go mod tidy
           echo ::set-output name=exit_code::$?
-      
+
```
nghazali, line 43: Do we need to add this in e2e tests? We already have in unit test workflow and we can skip here.

nghazali:
All good. To avoid failed tests in main, we can merge this ticket after TFKUBE-222

nghazali (approved):

jjeongatl (approved):

jjeongatl:
Should we fix the unit test?

nghazali (approved):

PR #109, opened 2022-01-18
TFKUBE-222: Cleanup e2e tests

nghazali:


test/e2etest/installer_test.go
```diff
@@ -3,31 +3,21 @@ package e2etest
...
-	product 	  = "bamboo"
-	domain  	  = "deplops.com"
 )
 
 func TestInstaller(t *testing.T) {
 
 	testConfig := createConfig(t)
 
+	// Schedule uninstall and cleanup the environment
+	defer runUninstallScript(testConfig.ConfigPath)
```
jjeongatl, line 14: Are we not separating this from install step?

nghazali: No need to do this. As we used `defer` this will run just before return regardless of the test result.

yzha645: I guess the reason we were separating them is to facilitate the troubleshooting when running locally. Keeping the infra there and uninstall whenever we finished debugging/testing/coding will save time. And in Github action, we always run both steps to avoid any leftover resources. 

For current implementation, do we have anything similar?

jjeongatl: I understand but the reason why we split this to another step was that sometimes when the test function unexpectedly exit(e.g. some runtime panic) defer didn't run. Maybe it is worth clarify/test this edge case

nghazali: Using 'defer' is placed at the begging of the test method before start installation and after creating the config. It means in case of any panic and disruption during installation/healthTest, uninstallation will run before return from `TestInstaller`.

jjeongatl (approved):

PR #110, opened 2022-01-20
TFKUBE-231: Made product version, bamboo and agent resources configurable

nghazali:


modules/products/bamboo/locals.tf
```diff
@@ -2,19 +2,20 @@ locals {
   product_name = "bamboo"
 
   helm_chart_repository     = "https://atlassian.github.io/data-center-helm-charts"
-  bamboo_helm_chart_version = "1.0.0"
-  agent_helm_chart_version  = "1.0.0"
+  bamboo_helm_chart_version = var.bamboo_configuration["helm_ver"]
```
yzha645, line 5: I think the complete form `helm_version` will be clearer.

nghazali: Thanks, is addressed.

yzha645 (approved):

PR #112, opened 2022-01-23
TFKUBE-234: Resume bamboo server after installation

nghazali:


install.sh
```diff
@@ -204,6 +204,20 @@ set_current_context_k8s() {
   fi
 }
 
+resume_bamboo_server() {
+  # Please note that if you import the dataset, make sure admin credential in config file (config.tfvars)
+  # is matched with admin info stored in dataset you import
+  ADMIN_USERNAME=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_admin_username' | sed -nE 's/^.*"(.*)".*$/\1/p')
+  ADMIN_PASSWORD=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep 'bamboo_admin_password' | sed -nE 's/^.*"(.*)".*$/\1/p')
```
nanux, line 211: We don't explicitly document that users should set the password when they are importing dataset and even if they do it can be potentially passed as a Terraform variable `TF_VAR_bamboo_admin_password` and not be in the config - we even [recommend that in our docs](https://atlassian-labs.github.io/data-center-terraform/userguide/CONFIGURATION/#sensitive-data).

I think we can possibly prompt for the password here?

install.sh
```diff
@@ -204,6 +204,20 @@ set_current_context_k8s() {
   fi
 }
 
+resume_bamboo_server() {
```
nanux, line 207: This should be called only when we are importing a dataset (the variable is set in config).

nanux (changes requested):

install.sh
nghazali: You are right. I changed this part to run with the following conditions:
* first run if the dataset url is provided
* get value from `TF_VAR_bamboo_admin_password` if `bamboo_admin_password` is not provided
* read the admin password from console if none of them are provided
* updated the documentation in config template to explicitly mention about the admin credential in case of providing the dataset

install.sh
nghazali: considered in change above

install.sh
```diff
@@ -207,10 +207,27 @@ set_current_context_k8s() {
...
+    fi
+    if [ -z "${ADMIN_PASSWORD}" ]
+      ADMIN_PASSWORD="${TF_VAR_bamboo_admin_password}"
+    fi
+    if [ -z "${ADMIN_USERNAME}" ]; then
+      read -p "Please enter the bamboo administrator username: " ADMIN_USERNAME
+    fi
+    if [ -n "${ADMIN_USERNAME}" ]; then
+      if [ -z "${ADMIN_PASSWORD}" ]; then
+        echo "Please enter the bamboo administrator password: "
```
nanux, line 228:
```suggestion
        echo "Please enter the Bamboo ${ADMIN_USERNAME} user password: "
```

nanux (approved):

PR #117, opened 2022-01-24
TFKUBE-234: Added the message in case of unsuccessful attempt resuming the server

nghazali:


install.sh
```diff
@@ -230,7 +231,14 @@ resume_bamboo_server() {
...
       bamboo_url=$(terraform output | grep '"bamboo" =' | sed -nE 's/^.*"(.*)".*$/\1/p')
       resume_bamboo_url="${bamboo_url}/rest/api/latest/server/resume"
-      curl -u "${ADMIN_USERNAME}:${ADMIN_PASSWORD}" -X POST "${resume_bamboo_url}"
+      local RESULT=$(curl -u "${ADMIN_USERNAME}:${ADMIN_PASSWORD}" -X POST "${resume_bamboo_url}")
+      if [[ "x${RESULT}" == *"RUNNING"* ]]; then
+        SERVER_STATUS="RUNNING"
+      fi
+    fi
+    if [ -z $SERVER_STATUS ]; then
+      log "We were not able to login into the Bamboo software to resume the server." "Warning"
```
nanux, line 240:
```suggestion
      log "We were not able to login into the Bamboo software to resume the server." "WARN"
```

install.sh
```diff
@@ -230,7 +231,14 @@ resume_bamboo_server() {
       fi
       bamboo_url=$(terraform output | grep '"bamboo" =' | sed -nE 's/^.*"(.*)".*$/\1/p')
       resume_bamboo_url="${bamboo_url}/rest/api/latest/server/resume"
-      curl -u "${ADMIN_USERNAME}:${ADMIN_PASSWORD}" -X POST "${resume_bamboo_url}"
+      local RESULT=$(curl -u "${ADMIN_USERNAME}:${ADMIN_PASSWORD}" -X POST "${resume_bamboo_url}")
+      if [[ "x${RESULT}" == *"RUNNING"* ]]; then
```
nanux, line 235: Should we retry or do you want to rely on the delay?

nghazali: Delay works consistently, but yes, retry is a better approach. I will change this part to retry with a shorter interval and a longer expiry time.

nghazali: However, retry is a better approach in that case. We ca change that part to retry with a shorter interval and a longer expiry time in a new ticket.

yzha645 (approved):

nanux (approved):

PR #121, opened 2022-01-27
TFKUBE-241: Added a function to fetch the variable and refactored the existing code

nghazali:


scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
   else
     echo "$(date '+%Y-%m-%d %H:%M:%S') [INFO] $1"
   fi
+}
+
+# Fetch the value of requested variable defined in config file
```
nanux, line 15:
```suggestion
# Fetch the value of the requested variable defined in Terraform config file
```

scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
...
+# params: $1 - variable name, $2 - config file full path
+get_variable(){
+  if [ "$#" -eq 2 ]; then
+    if [ ! -f "${2}" ]; then
+      echo "File ${2} is not existed."
+      return 1
+    fi
+    local VALUE=$(grep -o '^[^#]*' "${2}" | grep "${1}" | sed 's/ //g' | grep "${1}=" | sed -nE 's/^.*"(.*)".*$/\1/p')
+    if [ ! $(echo "${VALUE}" | wc -l) -eq 1 ];then
+      echo "ERROR - '${1}' is re-defined in '${2}'";
```
nanux, line 25:
```suggestion
      log "ERROR - '${1}' is re-defined in '${2}'" "ERROR";
```

scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
...
+      echo "${VALUE}"
```
nanux, line 26: Does this print both values?

scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
   else
     echo "$(date '+%Y-%m-%d %H:%M:%S') [INFO] $1"
   fi
+}
+
+# Fetch the value of requested variable defined in config file
+# params: $1 - variable name, $2 - config file full path
+get_variable(){
+  if [ "$#" -eq 2 ]; then
+    if [ ! -f "${2}" ]; then
```
nanux, line 19: I would prefer to define variables instead of using the `${1}` and `${2}`.

```
local var_name=${1}
local config_path=${2}
```

scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
...
+      echo "File ${2} is not existed."
```
nanux, line 20:
```suggestion
      echo "File ${2} doesn't exist."
```

scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
...
+      return 1
+    fi
+    local VALUE=$(grep -o '^[^#]*' "${2}" | grep "${1}" | sed 's/ //g' | grep "${1}=" | sed -nE 's/^.*"(.*)".*$/\1/p')
+    if [ ! $(echo "${VALUE}" | wc -l) -eq 1 ];then
```
nanux, line 24:
```suggestion
    if [ ! $(echo "${VALUE}" | wc -l) -eq 1 ]; then
```

scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
...
+      echo "ERROR - '${1}' is re-defined in '${2}'";
+      echo "${VALUE}"
+      return 1;
+    fi
+    echo "${VALUE}"
```
nanux, line 29:
```suggestion
    log "${1}: ${VALUE}"
```

scripts/common.sh
nanux, line 26:
```suggestion
      log "${VALUE}"
```

scripts/common.sh
```diff
@@ -10,4 +10,25 @@ log(){
...
+    return 0
+  fi
+  echo "Usage: fetch <config file> <variable name>"
```
nanux, line 32:
```suggestion
  echo "Usage: get_variable <config file> <variable name>"
```

nanux (changes requested):

scripts/common.sh
nghazali: yes, to show the multiple definition, but I think is not needed, is enough to let the user know the variable is multiple defined. 

I removed this line.

scripts/common.sh
nghazali: We should not put in log this line. This is actually the output of the function. This should be the absolute value of the given variable in the config file and this will be used by the caller.

scripts/common.sh
nghazali: Thanks for the catch. This was the initial name for the method and I changed it later. Also the parameters are in reverse order here. I fixed both issues.

nanux (approved):

PR #124, opened 2022-01-27
TFKUBE-245: updated config and docs

jjeongatl:


config.tfvars
```diff
@@ -35,7 +35,7 @@ desired_capacity = 2
 # https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html
 # https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#USER_PIOPS
 db_instance_class    = "db.t3.micro"
-db_allocated_storage = 1000
+db_allocated_storage = 100
```
nanux, line 38: Good catch. I think it can be even smaller if it is `db.t3.micro` by default.

nanux (approved):

nghazali (approved):

PR #137, opened 2022-02-02
TFKUBE-255: Avoid run e2e test when is not needed

nghazali:


.github/workflows/e2e-test.yaml
```diff
@@ -6,6 +6,7 @@ on:
     paths-ignore: # https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths
       - 'docs/**'
       - '.atlassian/**'
+      - 'test/**'
```
nanux, line 9:
```suggestion
      - 'test/unittest**'
```

If we have `test/**` it means the e2e test won't run if there is a change to the e2e test (which we want).

nanux (changes requested):

PR #138, opened 2022-02-02
Add a feature to install local helm charts 

nghazali:
In order to implement end2end test for helm charts, we need to use terraform install helm chart using the latest code in main branch of the data-cente-helm-charts repository. 
This PR introduces a new variable to get the path to the local bamboo helm chart which by default is null. This variable is for internal use and no documentation will be provided. If this variable fills with valid path to the helm chart, will override the helm chart version, repository, and name.


modules/products/bamboo/locals.tf
```diff
@@ -1,12 +1,20 @@
...
+  # Install local helm charts if local helm chart path is provided (for test purposes)
+  local_helm_charts_path = var.local_helm_charts_path == null ? "" : "${var.local_helm_charts_path}/src/main/charts"
+  use_local              = fileexists("${local.local_helm_charts_path}/${product_name}/Chart.yaml")
 
-  bamboo_software_resources = {
+  helm_chart_repository     = local.use_local ? null : "https://atlassian.github.io/data-center-helm-charts"
+  bamboo_helm_chart_name    = local.use_local ? "${local.local_helm_charts_path}/${product_name}" : local.product_name
+  bamboo_helm_chart_version = local.use_local ? null : var.bamboo_configuration["helm_version"]
+
+  agent_helm_chart_name    = local.use_local ? "${local.local_helm_charts_path}/${agent_name}" : local.agent_name
```
nanux, line 13: We don't check for the existence of the local `bamboo-agent` chart similar to the `use_local` variable. Should we check for it?

variables.tf
```diff
@@ -103,6 +103,12 @@ variable "bamboo_admin_email_address" {
   type        = string
 }
 
+variable "local_helm_charts_path" {
+  description = "Path to local Helm charts repo to install local helm charts"
```
nanux, line 107:
```suggestion
  description = "Path to a local directory with Helm charts to install"
```

dc-infrastructure.tf
```diff
@@ -45,4 +45,6 @@ module "bamboo" {
     "mem"          = var.bamboo_agent_mem
     "agent_count"  = var.number_of_bamboo_agents
   }
+
+  internal_use_e2e_path = var.internal_use_e2e_path
```
nanux, line 49:
```suggestion
  local_helm_charts_path = var.local_helm_charts_path
```

nanux (changes requested):

modules/products/bamboo/variables.tf
```diff
@@ -103,3 +103,9 @@ variable "bamboo_agent_configuration" {
     error_message = "Bamboo Agent configuration is not valid."
   }
 }
+
+variable "local_helm_charts_path" {
+  description = "Path to local Helm charts repo to install local helm charts"
```
nghazali, line 108:
```suggestion
  description = "Path to a local directory with Helm charts to install"
```

modules/products/bamboo/locals.tf
nghazali: Here this is my design:

We introduce this feature for e2e test. So, all Helm charts either pick from remote registry or local machine. so we have just one variable to define the root folder of `data-center-helm-charts` project. This check is for making sure the provided path is related to `data-center-helm-charts` project, and as we confirm the expected file is there, here we assumed the complete project is existed including all helm charts. 

But, your point is valid, we can add another check to make sure `bamboo-agent` is there, and this will cover an unusual use-case when user cloned the helm chart project and then deleted some folders/files (including bamboo agent chart file) or made a uncompleted copy of the project just for some products including bamboo and excluding bamboo agent.  (note we add `/src/main/charts` and then add `</<product>/Chart.yaml` to the provided local folder to make sure we have the root folder of the specific project [helm chart] to proceed).

nghazali: I refactored the code and made ability to use local Helm charts individually for each chart. In this case we can define to install any selective combinations for using local or remote helm chart.

dc-infrastructure.tf
```diff
@@ -45,4 +45,8 @@ module "bamboo" {
     "mem"          = var.bamboo_agent_mem
     "agent_count"  = var.number_of_bamboo_agents
   }
-}
\ No newline at end of file
+
+  # If local Helm charts is provided then terraform install charts locally and will ignore using remote registery
```
nghazali, line 49:
```suggestion
  # If local Helm charts path is provided then terraform installs charts locally and will ignore using remote registry
```

dc-infrastructure.tf
```diff
@@ -45,4 +45,8 @@ module "bamboo" {
     "mem"          = var.bamboo_agent_mem
     "agent_count"  = var.number_of_bamboo_agents
   }
-}
\ No newline at end of file
+
+  # If local Helm charts path is provided then terraform installs charts locally and will ignore using remote registry
```
yzha645, line 49:
```suggestion
  # If local Helm charts path is provided, Terraform will then install using local charts and ignores remote registry
```

yzha645 (approved):

nanux (approved):

PR #143, opened 2022-02-09
TFKUBE-302: Organised the config file and variables - added product list in config and install the bamboo selectively

nghazali:


variables.tf
```diff
@@ -170,3 +188,9 @@ variable "number_of_bamboo_agents" {
     error_message = "Number of agents must be greater than or equal to 0."
   }
 }
+
+variable "bamboo_install_local_chart" {
+  description = "Install Bamboo and Agents using local Helm charts"
```
nanux, line 193:
```suggestion
  description = "If true installs Bamboo and Agents using local Helm charts located in local_helm_charts_path"
```

outputs.tf
```diff
@@ -28,31 +38,21 @@ output "ingress" {
...
-output "product_urls" {
-  description = "URLs to access the deployed Atlassian products"
-
-  value = {
-    bamboo = module.bamboo.product_domain_name
-  }
-}
-
-output "database" {
+output "bamboo-database" {
```
nanux, line 41:
```suggestion
output "bamboo_database" {
```

I think we are using `_` underscore everywhere.

outputs.tf
```diff
@@ -28,31 +38,21 @@ output "ingress" {
...
   description = "Database information"
 
-  value = {
+  value = local.install_bamboo ? {
     rds_instance_id        = module.bamboo.rds_instance_id
     db_name                = module.bamboo.db_name
     kubernetes_secret_name = module.bamboo.kubernetes_rds_secret_name
     jdbc_connection        = module.bamboo.rds_jdbc_connection
-  }
+  } : null
```
nanux, line 49: Will this mean the value won't be included in the output at all?

config.tfvars
```diff
@@ -48,7 +32,7 @@ bamboo_admin_email_address = "<EMAIL ADDRESS>"
 #dataset_url = "https://centaurus-datasets.s3.amazonaws.com/bamboo/dcapt-bamboo.zip"
 
 
-# Custom tags for all resources to be created. Please add all tags you need to propagate among the resources.
+# Custom tags for all resources to be created. Please add all tags you need to propagate among the resources. (optional)
```
nanux, line 35:
```suggestion
# (optional) Custom tags for all resources to be created. Please add all tags you need to propagate among the resources.
```

config.tfvars
```diff
@@ -13,26 +13,10 @@ environment_name = "<ENVIRONMENT>"
...
-bamboo_admin_display_name  = "<DISPLAY NAME>"
-bamboo_admin_email_address = "<EMAIL ADDRESS>"
 
-
-################################################################################
-# Advanced
-################################################################################
+# List of the products to be installed. (optional) e.g. products = ["jira", "confluence"]
+# Supported products are jira, confluence, bitbucket, and bamboo.
+products = ["<LIST_OF_PRODUCTS>"]
```
nanux, line 19:
```suggestion
# (optional) List of the products to be installed.
# Supported products are jira, confluence, bitbucket, and bamboo.
# e.g.: products = ["jira", "confluence"]
products = ["<LIST_OF_PRODUCTS>"]
```

outputs.tf
nghazali: In case of `install_bamboo` is `false` it will print output like this:
```
bamboo_database: null
```

But prevent the error in installation, because `module.bamboo` does not existed.

nanux (approved):

PR #146, opened 2022-02-10
TFKUBE-270: Add Confluence installation

nanux:
Add Confluence installation and enable collaborative editing if the domain is specified.

- [x] add description for PR
- [x] check the e2e test - https://github.com/atlassian-labs/data-center-terraform/actions/runs/1827823637
- [x] check the unit tests


test/unittest/confluence_test.go
```diff
@@ -0,0 +1,85 @@
...
+	// verify Bamboo
+	confluenceKey := "helm_release.confluence"
+	terraform.RequirePlannedValuesMapKeyExists(t, plan, confluenceKey)
+	confluence := plan.ResourcePlannedValuesMap[confluenceKey]
+	assert.Equal(t, "deployed", confluence.AttributeValues["status"])
+	assert.Equal(t, "confluence", confluence.AttributeValues["chart"])
+	assert.Equal(t, "https://atlassian.github.io/data-center-helm-charts", confluence.AttributeValues["repository"])
+}
+
+func TestConfluenceSynchronyInstalled(t *testing.T) {
```
nanux, line 26: - [ ] This needs to be removed or fixed before merging.

config.tfvars
```diff
@@ -41,10 +35,71 @@ instance_types = ["m5.2xlarge"]
 # Desired number of nodes that the node group should launch with initially.
 desired_capacity = 1
 
+################################################################################
+# Jira Settings
```
errcode1202, line 39: Being picky here. Can these product specific config sections be ordered alphabetically.

test/e2etest/confluence_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+)
+
+func confluenceHealthTests(t *testing.T, testConfig TestConfig) {
+
+	// Test the RUNNING status
+	assertConfluenceStatusEndpoint(t, testConfig, "RUNNING")
+}
+
+// TODO: Add MORE integration tests here if needed
```
errcode1202, line 15: I think this comment should be removed.

test/e2etest/helper.go
```diff
@@ -80,16 +88,35 @@ func sendPostRequest(t *testing.T, url string, contentType string, body io.Reade
...
+
 	testConfig := TestConfig{
-		AwsRegion:       GetAvailableRegion(t),
-		BambooLicense:   bambooLicense,
-		EnvironmentName: EnvironmentName(),
-		ResourceOwner:   resourceOwner,
+		AwsRegion:         GetAvailableRegion(t),
+		EnvironmentName:   EnvironmentName(),
+		ResourceOwner:     resourceOwner,
+		JiraLicense:       getLicense(productList, "jira"),
```
errcode1202, line 116: How does this work if Jira does not support license pre-seeding?

errcode1202 (changes requested):

config.tfvars
nghazali: The reason behind this order is based on the number of users (bamboo has very low number of users so it goes to end). I think the most frequent used products go on top to be more accessible by the most of users. @errcode1202  WDYT?

test/e2etest/helper.go
nghazali: This is in test code, and this will not break things. All the logic is setting an environment variable named `<product>_license`.  As far as Jira does not support license pre-seeding, the `helper` function does not produce the environment variable for jira license, or even if it define that variable, jira does not pick it.  
I think even is better if we keep it for just in case if in future jira supports license pre-seeding.

errcode1202: If this just doesn't work (I dont see this being supported by Jira anytime soon) then why keep it? Come to think of it, this is even more a reason to get rid of it... Less code to maintain and all that 😄  "Support" for license pre-seeding is also being removed from the Jira Helm Charts. I just think keeping this around is going to make things confusing down the line.

config.tfvars
errcode1202: I didnt know that - Fair enough.

test/e2etest/helper.go
nghazali: fair enough :)

nghazali: Cleaned the code and removed jira license process.

modules/products/confluence/variables.tf
```diff
@@ -0,0 +1,90 @@
...
+  description = "Path to local Helm charts to install local confluence software"
+  type        = string
+  default     = ""
+  validation {
+    condition     = can(regex("^[.?\\/?[a-zA-Z0-9|\\-|_]*]*$", var.local_confluence_chart_path))
+    error_message = "Invalid local confluence Helm chart path."
+  }
+}
+
+variable "pvc_claim_name" {
```
yzha645, line 78: `pvc_name` or `persistent_volume_claim_name`. 

And why we're not passing in PVC object itself, but just the name?

test/e2etest/bamboo_test.go
```diff
@@ -46,17 +46,17 @@ func assertBambooProjects(t *testing.T, testConfig TestConfig) {
...
 	assert.Contains(t, string(content), "totalRecords: 1")
-	println("assertBambooProjects ..... PASSED")
+	println("assert Bamboo BambooProjects ..... PASSED")
 }
 
 func assertRemoteAgentList(t *testing.T, testConfig TestConfig) {
 	agentUrl := "admin/agent/configureAgents!doDefault.action"
 	// Wait 15 seconds to allow remote agents get online
-	time.Sleep(15*1000*time.Millisecond)
+	time.Sleep(15 * 1000 * time.Millisecond)
```
yzha645, line 55:
```suggestion
	time.Sleep(15 * time.Second)
```

modules/products/confluence/variables.tf
nghazali: I think `pvc` is a well known term in K8s and we don't get much benefit of using a long name here. 
The reason we don't pass whole module is we don't need the module. Only we need a valid 'pvc' name. We pass 'eks' and 'vpc' because to create RDS we need those modules.

nanux: I agree that `pvc_claim_name` is a little bit like `LCD Display` (Liquid Crystal Display Display). I don't have strong opinion here, I am happy to make it more explicit but I also see the `pvc` as a qualifier and `claim_name` as an attribute - making it `pvc_claim_name`.

yzha645 (approved):

nghazali (approved):

errcode1202 (approved):

PR #152, opened 2022-02-14
TFKUBE-308: Update documentation to take into account additional DC products

errcode1202:
Updates include:

- [x] Jira documentation
- [x] Confluence documentation
- [x] Bamboo documentation
- [x] Architectural diagrams for each product
- [x] Updates to minimum supported product versions


docs/docs/README.md
```diff
@@ -19,9 +19,26 @@ This project can be used for bootstrapping Atlassian Bamboo DC in a K8s cluster.
 
 ## Deployment overview
 
-The diagram below provides a high level overview of what a typical deployment will look like:
+The diagram below provides a high level overview of what a typical deployment will look like for each DC product:
+
```
nghazali, line 23: The product order in variable and config file is Jira, Confluence, and Bamboo. So is good to follow this order in documentation as well.

nghazali:
As we can have all products in the same cluster, does it worth to have a single diagram to show all products live in the same VPC and EKS rather than having separated diagram for each product?

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -5,7 +5,8 @@ In order to provision the infrastructure and install an Atlassian Data Center pr
 The content of the configuration file is divided into two groups:
 
 1. [Mandatory configuration](#mandatory-configuration)
```
nghazali, line 7: I think we should change this category fundementally. This was designed when we had just one product. Now we have three products and the new category should be :

1. Common configuration
2. Product specific Configuration

   a. Jira Configuration

   b. Confluence Configuration

   c. Bitbucket Configuration

   d. Bamboo Configuration
   

Obviousely there are some optional and mandetory settings in each group and those could be define by adding `(optional)` when we introduce the setting. See `config.tfvars`.

docs/docs/userguide/CONFIGURATION.md
```diff
@@ -64,58 +65,139 @@ region = "<REGION>"  # e.g. "ap-northeast-2"
 
     The value must be a valid [AWS region](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html){.external}.
 
+## Product specific configuration
```
nghazali, line 69: I think we missed database settings

docs/docs/README.md
errcode1202: Not sure I see a need for this following the ordering of whats in `tfvars.config`. From a documentation standpoint I think alphabetic makes more sense.

docs/docs/userguide/CONFIGURATION.md
errcode1202: Are you saying `Mandatory configuration` should be renamed to `Common configuration`? Additionally isnt the current optional config universal to all products hence it being in a section of its own?

docs/docs/userguide/CONFIGURATION.md
errcode1202: DB settings are under Optional configuration - not something I changed.

errcode1202:
> As we can have all products in the same cluster, does it worth to have a single diagram to show all products live in the same VPC and EKS rather than having separated diagram for each product?



> As we can have all products in the same cluster, does it worth to have a single diagram to show all products live in the same VPC and EKS rather than having separated diagram for each product?

This we discussed after standup. We felt it better to have a separate diagram for each type of deployment because depending on the product there will be different resources (ES/NFS) deployed. It was also felt trying to depict everything in one diagram would make it too busy. @nghazali what you think?

docs/docs/README.md
```diff
@@ -19,9 +19,29 @@ This project can be used for bootstrapping Atlassian Bamboo DC in a K8s cluster.
...
+=== "Jira"
+
+    !!! info "Architectural overview for Jira"
+
+    ![architecture](assets/images/jira/Terraform.png "Deployment")
+
+!!! tip "Multiple deployments to a single cluster"
+
+    Multiple DC products can also provisioned to the same cluster. See the [Configuration guide](./userguide/CONFIGURATION.md#products) for more details
```
yzha645, line 44:
```suggestion
    Multiple DC products can also be provisioned to the same cluster. See the [Configuration guide](./userguide/CONFIGURATION.md#products) for more details
```

docs/docs/README.md
```diff
@@ -33,9 +53,9 @@ The diagram below provides a high level overview of what a typical deployment wi
...
-| [8.1](https://confluence.atlassian.com/bamboo/bamboo-8-1-release-notes-1103070461.html){.external}  |
+|  Bamboo DC                                                                                         | Confluence DC                                                                                          | Jira DC                                                                                                                  |
+|----------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------|
+| [8.1](https://confluence.atlassian.com/bamboo/bamboo-8-1-release-notes-1103070461.html){.external} | [7.13](https://confluence.atlassian.com/doc/confluence-7-13-release-notes-1044114085.html){.external}  | [8.19](https://confluence.atlassian.com/jirasoftware/jira-software-8-19-x-release-notes-1082526044.html){.external}      |
```
yzha645, line 58: Do we want to use the LTS for Jira? i.e. 8.20.x.

errcode1202: We're reflecting what is [documented in the data-center-helm-charts ](https://atlassian.github.io/data-center-helm-charts/#product-versions) i.e. "The minimum versions that we support for each product are:" This makes me wonder, should the heading for this section be updated from:

`Atlassian Data Center products that can be provisioned with this tool:`

to

`The minimum versions that we support for each product are:`

docs/docs/userguide/CONFIGURATION.md
nghazali: As we have multi-product, the mandatory and optional concept is not making sense now. We have _common configuration_ and _product specific configuration_. In each category some items are mandatory (have no default value) and some are optional (have default value). Refactoring the documentation in this section is needed in my opinion.

docs/docs/userguide/CONFIGURATION.md
nghazali: This is the thing, we should get rid of `optional` / `mandatory` and aggregate all product related configuration.

nghazali:
> > As we can have all products in the same cluster, does it worth to have a single diagram to show all products live in the same VPC and EKS rather than having separated diagram for each product?
> 
> > As we can have all products in the same cluster, does it worth to have a single diagram to show all products live in the same VPC and EKS rather than having separated diagram for each product?
> 
> This we discussed after standup. We felt it better to have a separate diagram for each type of deployment because depending on the product there will be different resources (ES/NFS) deployed. It was also felt trying to depict everything in one diagram would make it too busy. @nghazali what you think?

How do we say that all product use the same VPC and the same EKS then (in architectural point of view)?

errcode1202:
> > > As we can have all products in the same cluster, does it worth to have a single diagram to show all products live in the same VPC and EKS rather than having separated diagram for each product?
> > 
> > 
> > > As we can have all products in the same cluster, does it worth to have a single diagram to show all products live in the same VPC and EKS rather than having separated diagram for each product?
> > 
> > 
> > This we discussed after standup. We felt it better to have a separate diagram for each type of deployment because depending on the product there will be different resources (ES/NFS) deployed. It was also felt trying to depict everything in one diagram would make it too busy. @nghazali what you think?
> 
> How do we say that all product use the same VPC and the same EKS then (in architectural point of view)?

Ive added this...

image

docs/docs/userguide/CONFIGURATION.md
errcode1202: I've updated things again - best viewed via `make docs`. Let me know what you think.

docs/docs/userguide/CONFIGURATION.md
errcode1202: Fixed.

nghazali (approved):

PR #154, opened 2022-02-15
TFKUBE-279: Initial Bitbucket helm chart support

yzha645:
This PR created the first version of Bitbucket module. 
Currently only RDS, ingress, license and resource configuration are included.


dc-infrastructure.tf
```diff
@@ -16,7 +16,6 @@ module "bamboo" {
   count      = local.install_bamboo ? 1 : 0
   depends_on = [module.base-infrastructure]
 
-  region_name      = var.region
```
nghazali, line 19: good catch :)

test/unittest/bitbucket_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"vpc": VpcDefaultModuleVariable,
+	//"pvc_claim_name":          "dummy_pvc_claimname",
+	"db_major_engine_version": "13",
+	"db_allocated_storage":    5,
+	"db_instance_class":       "dummy_db_instance_class",
+	"db_iops":                 1000,
+	//"admin_username":          "dummy_admin_username",
+	//"admin_password":          "dummy_admin_password",
+	//"admin_display_name":      "dummy_admin_display_name",
+	//"admin_email_address":     "dummy_admin_email_address",
```
nghazali, line 46: We should delete these comments

modules/products/bitbucket/helm_release.tf
```diff
@@ -0,0 +1,64 @@
...
+      #        }
+      #        sharedHome = {
+      #          customVolume = {
+      #            persistentVolumeClaim = {
+      #              claimName = var.pvc_claim_name
+      #            }
+      #          }
+      #          subPath = local.product_name
+      #        }
+      #      }
```
nghazali, line 51: I say let's uncomment this block and also `pvc_claim_name` variable, we will replace it with proper code when we add NFS later. 

Or delete the commented part from the code as it is not making sense to keep them.

variables.tf
```diff
@@ -215,6 +215,71 @@ variable "confluence_enable_synchrony" {
   default     = true
 }
 
+################################################################################
+# Bitbucket Settings
```
nghazali, line 219:
```suggestion
# Bitbucket Variables
```

nghazali (changes requested):

test/unittest/bitbucket_test.go
errcode1202: I cant remember from the conversation we had, but was there a reason to comment these out? BB supports credential pre-seeding so these should "just" work? If not Is there a ticket to re-instate admin credential pre-seeding?

modules/products/bitbucket/helm_release.tf
errcode1202: I agree, if we can keep these commented in with no impact on the deployment lets just keep them

errcode1202 (approved):

test/unittest/bitbucket_test.go
nghazali: I though we mentioned BB does not support system admin pre-seeding. If it does, so make sense to uncomment this part and make sure it works before merge the PR.

errcode1202: BB can be fully pre-seeded which includes admin creds: https://github.com/atlassian/data-center-helm-charts/blob/419a67465de0f62ec9f81b3b71223d62b189274c/src/main/charts/bitbucket/values.yaml#L540

nghazali: In this case we should not comment or remove those lines. We should add system admin credential variables for BB in root, and pass them into the BB module and apply in helm_release, the same as we have for Bamboo.

yzha645: yes, we should. But the conversation we had was to merge this PR first, and I can work on sysAdmin afterwards, so that other stories can be unblocked.

modules/products/bitbucket/helm_release.tf
yzha645: Updated.

config.tfvars
```diff
@@ -92,6 +92,35 @@ confluence_db_iops                 = 1000
...
+
+# Helm chart version of Bitbucket
+bitbucket_helm_chart_version = "1.2.0"
+
+# Bitbucket license
+# To avoid storing license in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_` (i.e. `TF_VAR_bitbucket_license`) and keep the below line commented out
+# If storing license as plain-text is not a concern for this environment, feel free to uncomment the following line and supply the license here
+#
+#bitbucket_license = "<LICENSE KEY>"
+
```
nghazali, line 107:
```suggestion

# Bitbucket system admin credentials
# WARNING: In case you are restoring an existing dataset (see the `dataset_url` property below), you will need to use credentials
# existing in the dataset to set this section. Otherwise any other value for the `bitbucket_admin_*` properties below are ignored.
# To avoid storing system admin password in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_` (i.e. `TF_VAR_bitbucket_admin_password`) and keep the below line commented out
# If storing password as plain-text is not a concern for this environment.

# To pre-seed the Bitbucket with the system admin information, feel free to uncomment the following settings and supply the system admin information:
#
#bitbucket_admin_username      = "<USERNAME>"
#bitbucket_admin_password      = "<PASSWORD>"
#bitbucket_admin_display_name  = "<DISPLAY NAME>"
#bitbucket_admin_email_address = "<EMAIL ADDRESS>"

```

test/unittest/bitbucket_test.go
yzha645: Added in sysadminCredentials.

modules/products/bitbucket/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+    condition     = can(regex("^[a-zA-Z]+[a-zA-Z0-9|\\-|_]*$", var.pvc_claim_name))
+    error_message = "Invalid claim name."
+  }
+}
+
+variable "admin_configuration" {
+  description = "Bitbucket admin configuration"
+  type        = map(any)
+  validation {
+    condition = (length(var.admin_configuration) == 4 &&
```
nanux, line 77: Does it mean it is now mandatory? It is probably fine to have it mandatory, I just want to highlight that Bitbucket doesn't have the same limitation as Bamboo (which needs to be fully configured if using unattended setup).

modules/products/bitbucket/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+  description = "Name of the environment."
+  type        = string
+  validation {
+    condition     = can(regex("^[a-zA-Z][a-zA-Z0-9\\-]{1,24}$", var.environment_name))
+    error_message = "Invalid environment name. Valid name is up to 25 characters starting with alphabet and followed by alphanumerics. '-' is allowed as well."
+  }
+}
+
+variable "namespace" {
+  description = "The namespace where Bitbucket pod will be installed."
```
nanux, line 11:
```suggestion
  description = "The namespace where Bitbucket helm chart will be installed."
```

modules/products/bitbucket/variables.tf
```diff
@@ -0,0 +1,81 @@
...
+  type        = string
+}
+
+variable "vpc" {
+  description = "VPC module that hosts the products."
+  type        = any
+}
+
+variable "eks" {
+  description = "EKS module that hosts the product."
```
nanux, line 21:
```suggestion
  description = "EKS module that hosts the products."
```

to align with `vpc` above (or we can use `product`)

modules/products/bitbucket/helm_release.tf
```diff
@@ -0,0 +1,67 @@
...
+          }
+          container = {
+            requests = {
+              cpu    = local.bitbucket_software_resources.cpu
+              memory = local.bitbucket_software_resources.mem
+            }
+          }
+        }
+        sysadminCredentials = {
+          secretName = kubernetes_secret.admin_secret.metadata[0].name
```
nanux, line 30: It doesn't need to be part of this PR but Bitbucket has the option to set up the DisplayName as well, it could be added as a variable with auto-generated default:

`bitbucket.displayName`

https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/bitbucket/values.yaml#L567

config.tfvars
```diff
@@ -92,6 +92,47 @@ confluence_db_iops                 = 1000
...
+# Bitbucket system admin credentials
+# To pre-seed Bitbucket with the system admin information, uncomment the following settings and supply the system admin information:
+#
+# To avoid storing password in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_`
+# (i.e. `TF_VAR_bitbucket_admin_password`) and keep `bamboo_admin_password` commented out
```
nanux, line 112:
```suggestion
# (i.e. `TF_VAR_bitbucket_admin_password`) and keep `bitbucket_admin_password` commented out
```

config.tfvars
```diff
@@ -92,6 +92,47 @@ confluence_db_iops                 = 1000
...
+# If storing password as plain-text is not a concern for this environment, feel free to uncomment `bamboo_admin_password` and supply system admin password here
```
nanux, line 113:
```suggestion
# If storing password as plain-text is not a concern for this environment, feel free to uncomment `bitbucket_admin_password` and supply system admin password here
```

nanux (changes requested):

modules/products/bitbucket/helm_release.tf
yzha645: When you say `auto-generated default`, do you mean something like `Bitbucket(<environment_name>)`?

modules/products/bitbucket/helm_release.tf
errcode1202: 🌮

modules/products/bitbucket/variables.tf
errcode1202: Good point!

nghazali: No, they are not mandatory, because the input values come from root level and all have default value there. The same for Bamboo.

modules/products/bitbucket/helm_release.tf
nanux: Yes, something like that. But you don't need to worry in this PR, maybe just adding a subtask to the user story.

modules/products/bitbucket/variables.tf
yzha645:
```suggestion
  description = "The namespace where Bitbucket Helm chart will be installed."
```

modules/products/bitbucket/variables.tf
yzha645: As discussed, created [TFKUBE-330](https://bulldog.internal.atlassian.com/browse/TFKUBE-330).

modules/products/bitbucket/helm_release.tf
yzha645: Created [TFKUBE-330](https://bulldog.internal.atlassian.com/browse/TFKUBE-330).

nanux (approved):

nghazali (approved):

PR #156, opened 2022-02-15
TFKUBE-322: Add support for collaborative editing if no domain is provided

nanux:
It is not perfect but I found this solution much easier simpler than trying to proactively create an elastic IP and then convince the new LoadBalancer to assign to it.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -208,12 +208,14 @@ set_current_context_k8s() {
 resume_bamboo_server() {
   # Please note that if you import the dataset, make sure admin credential in config file (config.tfvars)
   # is matched with admin info stored in dataset you import. 
-  log "Resuming Bamboo server."
   BAMBOO_DATASET=$(get_variable 'dataset_url' "${CONFIG_ABS_PATH}")
+  INSTALLED_BAMBOO=$(get_variable 'products' "${CONFIG_ABS_PATH}" | grep 'bamboo')
   local SERVER_STATUS=
 
   # resume the server only if a dataset is imported
-  if [ -n "${BAMBOO_DATASET}" ]; then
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -208,12 +208,14 @@ set_current_context_k8s() {
...
-  log "Resuming Bamboo server."
   BAMBOO_DATASET=$(get_variable 'dataset_url' "${CONFIG_ABS_PATH}")
+  INSTALLED_BAMBOO=$(get_variable 'products' "${CONFIG_ABS_PATH}" | grep 'bamboo')
   local SERVER_STATUS=
 
   # resume the server only if a dataset is imported
-  if [ -n "${BAMBOO_DATASET}" ]; then
+  if [ -n "${BAMBOO_DATASET}" ] && [ -n "${INSTALLED_BAMBOO}" ]; then
+    log "Resuming Bamboo server."
+
```
nanux, line 218: The message should be printed only if we are actually doing it.

install.sh
```diff
@@ -270,5 +284,8 @@ add_tags_to_asg_resources
 # Resume bamboo server if the credential is provided
 resume_bamboo_server
 
+# Set the correct Synchrony URL
+set_synchrony_url
```
nanux, line 288: The output of this function if Confluence was installed and domain is not provided:

```
2022-02-15 12:26:28 [INFO] Configuring the Synchrony service.
2022-02-15 12:26:48 [INFO] Synchrony URL is set to 'http://ac8230a53e6ad4bacb27da6028483ec5-1351588498.ap-southeast-1.elb.amazonaws.com/synchrony'.
```

nghazali (approved):
All good 👍

PR #157, opened 2022-02-15
TFKUBE-306: Add confluence_enable_synchrony to docs

nanux:
Almost all the work was done by @errcode1202 in a previous PR, kudos! 🙇


docs/docs/userguide/CONFIGURATION.md
```diff
@@ -338,7 +338,15 @@ desired_capacity = <NUMBER OF NODES>  # between 1 and 10
     confluence_min_heap            = "256m"
     confluence_max_heap            = "512m"
     ```
+    
+    ### Collaborative editing
+
+    `confluence_enable_synchrony` enables [Collaborative editing](https://confluence.atlassian.com/doc/collaborative-editing-858771779.html). (default: `true`)
 
+    ```terraform
+    confluence_enable_synchrony = true
```
nanux, line 347: When I was adding the property, shouldn't we call it `confluence_enabled_collaborative_editing`? I am not sure if Synchrony is a commonly known term.

errcode1202: I agree I think `confluence_enabled_collaborative_editing` is a better fit

nanux:
@errcode1202 the shortcoming of using the tabs for the product configuration is that if you have selected a specific product (e.g. Bamboo) and you will click on the property from a different product (e.g. Confluence License) in the right sidebar, it doesn't scroll to the property. You need to select the correct tab to scroll to the right place.
Screen Shot 2022-02-15 at 5 13 37 pm

errcode1202:
> @errcode1202 the shortcoming of using the tabs for the product configuration is that if you have selected a specific product (e.g. Bamboo) and you will click on the property from a different product (e.g. Confluence License) in the right sidebar, it doesn't scroll to the property. You need to select the correct tab to scroll to the right place. Screen Shot 2022-02-15 at 5 13 37 pm

@nanux thats annoying I guess the `pymdownx.tabbed` extension doesn't support that behaviour. Do you think there is a tradeoff we need to make here between aesthetics and navigation?

errcode1202:
@nanux to address this Ive created this PR https://github.com/atlassian-labs/data-center-terraform/pull/158 as discussed

docs/docs/userguide/CONFIGURATION.md
nghazali: I think even `confluence_collaborative-editing_enabled` is better option.

nghazali (approved):

errcode1202 (approved):

docs/docs/userguide/CONFIGURATION.md
nanux: I have renamed the property to `confluence_collaborative_editing_enabled`.

PR #160, opened 2022-02-16
TFKUBE-332: Made system admin credential optional for Bitbucket.

nghazali:
Also the following actions are included in this PR:
* Ignored admin credential information for bamboo if the Bamboo is not installed, 
* Fixed fetching the product name in scripts (`products` is a list and `get_variable` does not work for the `list`)
* Renamed `helm_release.tf` to `helm.tf` for Bitbucket and Jira
* Fixed the unit tests
* Updated the Helm chart version in config template to `1.2.0` for all products


install.sh
```diff
@@ -106,18 +102,24 @@ verify_configuration_file() {
...
+        HAS_VALIDATION_ERR=1
+      fi
     fi
-  fi
 
-  if [ -z "${POPULATED_ADMIN_PWD}" ];  then
-    if [ -z "${TF_VAR_bamboo_admin_password}" ]; then
-      log "Admin password is missing. Please provide admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'." "ERROR"
-      HAS_VALIDATION_ERR=1
+    if [ -z "${POPULATED_ADMIN_PWD}" ];  then
```
errcode1202, line 118:
```suggestion
    if [ -z "${POPULATED_ADMIN_PWD}" ]  && [ -z "${TF_VAR_bamboo_admin_password}" ]
then
   log "Admin password is missing. Please provide Bamboo admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'." "ERROR"
        HAS_VALIDATION_ERR=1
fi
```

errcode1202, line 118: Can these nested `if` blocks be simplified like so

```shell
if [ -z "${POPULATED_ADMIN_PWD}" ]  && [ -z "${TF_VAR_bamboo_admin_password}" ]
then
   log "Admin password is missing. Please provide Bamboo admin password in config file, or export it to the environment variable 'TF_VAR_bamboo_admin_password'." "ERROR"
        HAS_VALIDATION_ERR=1
fi
```

nghazali: Thanks for pointing this. Yes, we definitely can simplify this.

Done

install.sh
```diff
@@ -106,17 +102,18 @@ verify_configuration_file() {
     log "${INVALID_CONTENT}"
     HAS_VALIDATION_ERR=1
   fi
-
-  if [ -z "${POPULATED_LICENSE}" ];  then
-    if [ -z "${TF_VAR_bamboo_license}" ]; then
-      log "License is missing. Please provide license in config file, or export it to the environment variable 'TF_VAR_bamboo_license'." "ERROR"
+  INSTALLED_BAMBOO=$(get_product "bamboo" "${CONFIG_ABS_PATH}")
```
errcode1202, line 105: I think this should be `INSTALL_BAMBOO`

install.sh
```diff
@@ -253,7 +250,7 @@ resume_bamboo_server() {
 
 set_synchrony_url() {
   DOMAIN=$(get_variable 'domain' "${CONFIG_ABS_PATH}")
-  INSTALLED_CONFLUENCE=$(get_variable 'products' "${CONFIG_ABS_PATH}" | grep 'confluence')
+  INSTALLED_CONFLUENCE=$(get_product "confluence" "${CONFIG_ABS_PATH}")
```
errcode1202, line 253: `INSTALL_CONFLUENCE`

modules/products/bitbucket/locals.tf
```diff
@@ -11,6 +11,14 @@ locals {
     "mem" : var.bitbucket_configuration["mem"]
   }
 
+  admin_settings = length(kubernetes_secret.admin_secret) == 1 ? yamlencode({
+    bitbucket = {
+      sysadminCredentials = {
+        secretName = kubernetes_secret.admin_secret[0].metadata[0].name
+      }
+    }
+  }) : ""
```
yzha645, line 20:
```suggestion
  }) : yamlencode({})
```

nghazali: Sorry, I saw this comment a bit late. I will apply this change in next PR.

PR #161, opened 2022-02-16
Add support for documentation annotations

errcode1202:
Add support for doing this...

image

### DIFF
--- a/docs/docs/userguide/configuration/CONFIGURATION.md
+++ b/docs/docs/userguide/configuration/CONFIGURATION.md
@@ -63,7 +63,6 @@ region = "<REGION>"  # e.g. "ap-northeast-2" # (1)
 
 1. A valid [AWS region](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html){.external} must be used.
 
-
 ### Products
 
 The `products` list can be configured with one or many products. This will result in these products being deployed to the same K8s cluster. For example, if a Jira and Confluence deployment is required this property can be configured as follows:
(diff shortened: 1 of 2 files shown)


docs/docs/userguide/configuration/CONFIGURATION.md
```diff
@@ -125,14 +126,12 @@ resource_tags = {
 `instance_types` defines the instance type for the EKS cluster node group.
 
 ```terraform
-instance_types = ["m5.2xlarge"]
+instance_types = ["m5.2xlarge"] # (1)
```
nghazali, line 129: I think for very important information we should display the warning without need of user click. We may want that users see this warning or any other important messages without hiding them behind a click request.

nanux: We need to land on a decision on this PR to align ourselves with regards to how to use this docs feature. From my perspective, I think we should hide information that clutters documentation but it can be easily deduced. For example, 

**I would hide under tooltip:**

* Region description - from the name of the variable and the example value, it is quite easy to guess what needs to be there
* Instance type value - I think we can hide the link to the instance types, again, something that is quite easy to deduce

**I wouldn't hide:**

* Warning regarding the static nature of the instance type - if you create a cluster with instance type, you shouldn't know it is a one-way door (at least for now). It shouldn't come as a surprise.
* Values available to the product list, this should be shown at a glance. Users cannot easily deduce it (it is our decision which products we support and forcing them to hover over it, we don't really gain much.

That is my perspective.

errcode1202:
Closing this it was only ever meant to be used as a guide on how annotations can be used.

PR #162, opened 2022-02-18
TFKUBE-283: NFS support for bitbucket

errcode1202:
- [x] Tag NFS helm chart `tgz` with version:
https://github.com/atlassian/data-center-helm-charts/blob/main/docs/docs/examples/storage/nfs/nfs-server-example-0.1.0.tgz


modules/products/bitbucket/nfs/main.tf
```diff
@@ -0,0 +1,22 @@
+resource "helm_release" "nfs" {
+  chart     = "https://raw.githubusercontent.com/atlassian/data-center-helm-charts/main/docs/docs/examples/storage/nfs/nfs-server-example.tgz"
```
nghazali, line 2: This chart is for just a sample and is not for production as is mentioned in the `Chart.yaml` description:
```
 name: nfs-server-example
description: A NFS server example using NFS Ganesha. Do not use in production.
```

Also there are another [stable chart](https://github.com/helm/charts/tree/master/stable/nfs-server-provisioner) for nfs provisioner in github but unfortunately is deprecated a couple of years ago and not logical to use it in production. 

We maybe can implement NFS feature without a helm chart using s3 bucket or any proper storage class, [storage gateway](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/storagegateway_gateway#s3-file-gateway) and [AWS Gateway NFS file share](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/storagegateway_nfs_file_share) instead as an alternative.

errcode1202: Some background on this Chart... When developing the the DC helm charts a lot of time was spent trying to answer the question of how we provide an NFS solution. With no obvious Helm chart offerings available (all of them appear to be deprecated - including a chart supplied by Google) we were a little reluctant to hand crank our own solution with the overhead of having to provide additional support for something that can be rather nebulous/granular i.e. an NFS server. This chart is based on quite a bit of investigation and ultimately resulted in NFS Ganesha being used under the hood with specific Bitbucket config. Although It works for its intended purpose, we don't want to be in the business of supporting NFS config hence the description and caveat on the chart. 

My concern with AWS Gateway NFS file share is that It's not something that I imagine the [Bitbucket team has tested from a functional/performance throughput standpoint](https://confluence.atlassian.com/bitbucketserver/install-bitbucket-data-center-872139817.html#InstallBitbucketDataCenter-nfsStep2.Provisionyoursharedfilesystem). 

We can certainly explore the AWS Gateway option, but ultimately I think we should take this to DCAPT and see how this compares to the expectations they will have on BB r/w performance. @nanux what are your thoughts on this?

errcode1202: Another thing to keep in mind is that this Terraform is not going to be officially supported. So that could feed into the decision making on whether using this NFS chart is acceptable or not.

nanux: In short, we don't currently have other options how to deploy the shared home for Bitbucket that would comply with its performance requirements. The fact that the helm chart is not officially supported shouldn't prevent us to use the chart in non-supported testing environments. We cannot just "replace" the shared home storage implementation (if that would be easily possible, it would have been done). In principle I don't see a problem with this implementatin for now (I will leave some comments regarding specific points).
The main problem with this line/approach is that we don't have any versioning for the helm chart. I think we should add a version number to the artifact so we can have full control over the installed content.

modules/products/bitbucket/nfs/variables.tf
```diff
@@ -0,0 +1,22 @@
+variable "product" {
```
nanux, line 1: How this is used or intended to be used? Is it really necessary? I think if we are using implicit dependency in the `helm_release` resource for the PVC, it should create the nfs module by default?

modules/products/bitbucket/nfs/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+  description = "The product that NFS will be created for."
+  type        = string
+}
+
+variable "namespace" {
+  description = "Kubernetes namespace to install Bitbucket."
```
nanux, line 7:
```suggestion
  description = "Kubernetes namespace to install NFS server."
```

Shouldn't we always install NFS to the namespace?

modules/products/bitbucket/nfs/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+  type        = string
+}
+
+variable "chart_name" {
+  description = "The chart name to use."
```
nanux, line 12: It is not name of the chart, is it? It is used in the `values` so I assume it changes the helm chart behaviour. Does it change the `helm release name`?

errcode1202: So we pass this to have control over the name of the release and resultant pod names. By default this would be `nfs-server-example`. Through this variable we override the name so that it is now `bitbucket-nfs-server`

errcode1202: It might not be necessary to do this though - maybe we can live with it being called `nfs-server-example` and hence remove the variable?

nanux: No, I think it is a good idea to override it.

nanux (approved):

modules/products/bitbucket/nfs/main.tf
nghazali: In this case why don't we use this [stable chart](https://github.com/helm/charts/tree/master/stable/nfs-server-provisioner)?

errcode1202: I think because it's been deprecated. In this case I'd argue its better to use our own, additionally our chart was specifically tuned for BB purposes.

PR #163, opened 2022-02-18
TFKUBE-286: Add elasticsearch to Bitbucket

nghazali:


modules/AWS/elasticsearch/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+  type        = string
+  default     = "r4.large.elasticsearch"
+  validation {
+    condition     = can(regex("^(m[3-5]|r[3-5]|c[4-5]|i[2-3]|d2|t2|ultrawarm1).(1[0|2|4|6|8]x|2x|4x|8x|9x|x)?large.elasticsearch$", var.instance_type))
+    error_message = "Elasticsearch instance type is invalid. see 'https://aws.amazon.com/ec2/instance-types'."
+  }
+}
+
+variable "instance_count" {
+  description = "Number of the elasticsearch instances."
```
errcode1202, line 26: You can drop, the word "the" from the description i.e. "Number of elasticsearch instances."

modules/AWS/elasticsearch/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+  description = "Name of the environment."
+  type        = string
+  validation {
+    condition     = can(regex("^[a-zA-Z][a-zA-Z0-9\\-]{1,24}$", var.environment_name))
+    error_message = "Invalid environment name. Valid name is up to 25 characters starting with alphabet and followed by alphanumerics. '-' is allowed as well."
+  }
+}
+
+variable "vpc_id" {
+  description = "VPC ID that hosts the elasticsearch."
```
errcode1202, line 11: Maybe something like:

"ID of the VPC that hosts the elasticsearch instance."

modules/AWS/rds/main.tf
```diff
@@ -23,7 +23,7 @@ module "security_group" {
 
 module "db" {
   source  = "terraform-aws-modules/rds/aws"
-  version = "~> 3.0"
+  version = "~> 3.5"
```
errcode1202, line 26: Any reason for the minor bump here - assuming its latest?

errcode1202: With the pessimistic operator I assume we're fine if versions `3.5` -> `3.9` can be used?

nghazali: The latest is 4.1.2, we had to upgrade the RDS version as there is a deprecated input causing a warning message when we run terraform:
```
Warning: Argument is deprecated

  with module.bamboo.module.database.module.db.module.db_instance.aws_db_instance.this,
  on .terraform/modules/bamboo.database.db/modules/db_instance/main.tf line 34, in resource "aws_db_instance" "this":
  34:   name                                = var.name

Use db_name instead

(and 9 more similar warnings elsewhere)
```

I just double checked and it seems we should upgrade the rds module to `4.0` because in `3.5` still this change is not applied.

nghazali: I just noticed we already have raised a ticket for this problem, so I revert this change as it will be addressed separately.

variables.tf
```diff
@@ -305,6 +305,60 @@ variable "bitbucket_max_heap" {
...
+  default     = "r4.large.elasticsearch"
+}
+
+variable "bitbucket_elasticsearch_replicas" {
+  description = "Number of nodes for elasticsearch instance"
+  type        = number
+  default     = 2
+  validation {
+    condition     = var.bitbucket_elasticsearch_replicas > 1 && var.bitbucket_elasticsearch_replicas < 7
+    error_message = "Number of elasticsearch nodes must be between 2 and 6."
```
nanux, line 358: Does it work with two nodes only? I thought the ES health check requires 3 nodes.

variables.tf
```diff
@@ -305,6 +305,60 @@ variable "bitbucket_max_heap" {
...
+variable "bitbucket_elasticsearch_storage" {
+  description = "Storage size for Bitbucket elasticsearch instance in GiB."
+  type        = number
+  default     = 10
+}
+
+variable "bitbucket_aws_elasticsearch_instance_type" {
+  description = "Instance type for Bitbucket AWS Elasticsearch."
+  type        = string
+  default     = "r4.large.elasticsearch"
```
nanux, line 349: Why the default is `r4.large.elasticsearch`? Sounds pretty old and large, comparing to the DB instance which is `db.t3.micro` by default. I would say DB should have the same or even more performant instance type than Elasticsearch.

variables.tf
nghazali: Yes, it works with minimum 2 nodes.

errcode1202 (approved):
🚀

PR #165, opened 2022-02-21
TFKUBE-335: Make Jira version configurable. 

nghazali:


variables.tf
```diff
@@ -85,6 +85,12 @@ variable "jira_helm_chart_version" {
   default     = "1.1.0"
 }
 
+variable "jira_tag_version" {
```
nanux, line 88:
```suggestion
variable "jira_version_tag" {
```

modules/products/jira/variables.tf
```diff
@@ -57,6 +57,12 @@ variable "jira_configuration" {
   }
 }
 
+variable "tag_version" {
```
nanux, line 60:
```suggestion
variable "version_tag" {
```

modules/products/jira/locals.tf
```diff
@@ -37,5 +37,11 @@ locals {
     }
   })
 
+  tag_version = var.tag_version != null ? yamlencode({
```
nanux, line 40:
```suggestion
  version_tag = var.version_tag != null ? yamlencode({
```

modules/products/jira/locals.tf
```diff
@@ -37,5 +37,11 @@ locals {
...
+    image = {
+      tag = var.tag_version
```
nanux, line 42:
```suggestion
      tag = var.version_tag
```

modules/products/jira/helm.tf
```diff
@@ -53,6 +53,7 @@ resource "helm_release" "jira" {
       }
     }),
     local.ingress_settings,
+    local.tag_version,
```
nanux, line 56:
```suggestion
    local.version_tag,
```

install.sh
```diff
@@ -257,6 +257,8 @@ set_synchrony_url() {
     SYNCHRONY_FULL_URL=$(terraform output | sed "s/ //g" | grep "synchrony_url=" | sed -nE 's/^.*"(.*)".*$/\1/p')
     helm upgrade confluence atlassian-data-center/confluence -n atlassian --reuse-values --set synchrony.ingressUrl="${SYNCHRONY_FULL_URL}" > /dev/null
     log "Synchrony URL is set to '${SYNCHRONY_FULL_URL}'."
+    echo
+    helm list -n atlassian
```
nanux, line 261: This will happen only if confluence is installed and doesn't have a domain. I think we can print the helm releases every time. I think it deserves its own function at the end of execution.

config.tfvars
```diff
@@ -42,6 +42,9 @@ desired_capacity = 1
 
 # Helm chart version of Jira
 jira_helm_chart_version = "1.2.0"
+# The version of Jira software is defaulted in Helm chart. If you wish to install a different
```
nanux, line 45:
```suggestion
# By default, Jira Software will use the version defined in the helm chart. If you wish to override the version, uncomment the following line and set the jira_software_version to any of the versions available on https://hub.docker.com/r/atlassian/jira-software/tags
```

nanux (changes requested):

config.tfvars
```diff
@@ -42,6 +42,9 @@ desired_capacity = 1
...
+# version of Jira software you can uncomment the following line and define your desired version.
+#jira_software_version   = "8.20.5-jdk11"
```
nanux, line 47: I am not sure if we should include default, even commented out. This means we need to keep it up to date as customers might uncomment the line thinking that this is the default (from helm chart). I think we can leave this empty. As we point them to the docker hub tags (comment above), they can see examples of the versions that are available.

install.sh
nghazali: Right, I moved this after if block.

errcode1202 (approved):

nanux (approved):

PR #166, opened 2022-02-22
TFKUBE-343: Fixed the synchrony url update for Confluence when domain…

nghazali:
… is not provided.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -258,6 +258,7 @@ set_synchrony_url() {
     helm upgrade confluence atlassian-data-center/confluence -n atlassian --reuse-values --set synchrony.ingressUrl="${SYNCHRONY_FULL_URL}" > /dev/null
     log "Synchrony URL is set to '${SYNCHRONY_FULL_URL}'."
   fi
+  helm list --namespace atlassian
 }
 
 # Process the arguments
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -258,6 +258,7 @@ set_synchrony_url() {
     helm upgrade confluence atlassian-data-center/confluence -n atlassian --reuse-values --set synchrony.ingressUrl="${SYNCHRONY_FULL_URL}" > /dev/null
     log "Synchrony URL is set to '${SYNCHRONY_FULL_URL}'."
   fi
+  helm list --namespace atlassian
```
nanux, line 261: Can we move it out of synchrony function?

install.sh
```diff
@@ -281,8 +282,9 @@ add_tags_to_asg_resources
 # Resume bamboo server if the credential is provided
 resume_bamboo_server
 
+# Print information about manually adding the new k8s context
+set_current_context_k8s
```
nanux, line 286: 🤦 sorry I made this change locally and thought that I have merged it. Thanks for fixing it!

nanux (approved):

install.sh
nghazali: Yes, moved it at the end of script.

PR #169, opened 2022-02-22
TFKUBE-281: Bitbucket e2e tests

errcode1202:
Passing on branch e2e test:

https://github.com/atlassian-labs/data-center-terraform/actions/runs/1879840860


test/e2etest/bitbucket_test.go
```diff
@@ -0,0 +1,48 @@
...
+	assertBitbucketStatusEndpoint(t, testConfig, "RUNNING")
+}
+
+func assertBitbucketStatusEndpoint(t *testing.T, testConfig TestConfig, expectedStatus string) {
+	statusUrl := "status"
+	url := fmt.Sprintf("https://%s.%s.%s/%s", bitbucket, testConfig.EnvironmentName, domain, statusUrl)
+	content := getPageContent(t, url)
+	println("Asserting Bitbucket Status Endpoint ...")
+	if assert.Contains(t, string(content), expectedStatus) {
+		assertNfsConnectivity(t, testConfig)
```
nghazali, line 20: This is a separated test and should not be called by status endpoint test. Would you please move this call to `bitbucketHealthTests` and an independent test after calling the status test?

test/e2etest/installer_test.go
```diff
@@ -9,12 +9,14 @@ import (
...
 
 	// List of the products to test
-	productList := []string{jira, confluence, bamboo}
+	productList := []string{jira, confluence, bamboo, bitbucket}
 	testConfig := createConfig(t, productList)
 
 	// Schedule uninstall and cleanup the environment
 	defer runUninstallScript(testConfig.ConfigPath)
 
+	println("################## AWS REGION " + testConfig.AwsRegion + " ##################")
```
nghazali, line 18:
```suggestion
	println(fmt.Sprintf("################## AWS REGION '%s' ##################", testConfig.AwsRegion))
```

test/e2etest/bitbucket_test.go
errcode1202: But if the status endpoint test fails then why bother calling the nfs connectivity test - this approach saves unnecessary calls and cpu cycles

errcode1202: The thinking behind this is, if the status endpoint test fails then that gives a good indication that the BB pod will be unavailable/down. If its not available then the NFS test will just not pass.

nghazali: I totally understand your reasoning behind this. However, this is the usual test pattern and better to follow it. We have the same approach in other products (at this point only Bitbucket has more than one test). 
About your concern, the provisioning the environment and destroying it takes more than 40 minutes, and all our tests will take not more than a few seconds. Although we know the test will ends to fail but still we have the log for conducting that test and failure result.

errcode1202: I can agree on following a pattern, and I'll make the change to follow it, but now Im questioning whether the current pattern is correct? Running all of the additional product tests (for instance bamboo project checks) in the face of a failing `status` test feels like a pointless exercise. Shouldn't our stance be, if the `status` is bad then skip any other associated product tests? Little improvements like this although small will reduce run time and eliminate waste... Think big start small ;) 

It's an interesting discussion, that we can maybe follow up on later.

nghazali: Yap, we can discuss this later. Even if we should stop the rest of tests in case of failure in status check, we need to add this checkpoint in product health check itself and not in status check.

modules/AWS/eks/main.tf
```diff
@@ -11,6 +11,9 @@ module "eks" {
   vpc_id  = var.vpc_id
   subnets = var.subnets
 
+  kubeconfig_aws_authenticator_command = "aws"
```
nanux, line 14: Can you please add the comment regarding the removal of these variables in eks module v 18? Ideally, link https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/UPGRADE-18.0.md

test/e2etest/bitbucket_test.go
```diff
@@ -0,0 +1,47 @@
...
+	print("Asserting Bitbucket NFS connectivity ...")
+	returnCode, kubectlError := k8s.RunKubectlAndGetOutputE(t, kubectlOptions,
+		"exec", "bitbucket-nfs-server-0",
+		"--", "/bin/bash",
+		"-c", "echo \"Greetings from an NFS\" >> /srv/nfs/nfs-file-share-test.txt; echo $?")
+
+	assert.Nil(t, kubectlError)
+	assert.Equal(t, "0", returnCode)
+
+	// Read the file from the Bitbucket pod
```
nanux, line 38: Nice test!

nghazali (approved):

test/e2etest/installer_test.go
```diff
@@ -9,12 +10,14 @@ import (
...
 
 	// List of the products to test
-	productList := []string{jira, confluence, bamboo}
+	productList := []string{jira, confluence, bamboo, bitbucket}
 	testConfig := createConfig(t, productList)
 
 	// Schedule uninstall and cleanup the environment
 	defer runUninstallScript(testConfig.ConfigPath)
 
+	println(fmt.Sprintf("################## AWS REGION '%s' ##################", testConfig.AwsRegion))
```
nanux, line 19: When we started doing our own test output formatting, can we put a similar line between each product assertion?

test/e2etest/installer_test.go
```diff
@@ -32,6 +35,11 @@ func TestInstaller(t *testing.T) {
 	if contains(productList, confluence) {
```
nanux, line 35:
```suggestion
	println(fmt.Sprintf("################## Confluence Tests ##################", testConfig.AwsRegion))
	if contains(productList, confluence) {
```

nanux (approved):

test/e2etest/installer_test.go
nghazali: Actually it needs a `%s` if we want to include `AwsRegion`, but in this case I think we can simple have:
```
println("################## Confluence Tests ##################")
```

test/e2etest/bitbucket_test.go
nanux: I think you both have valid points (and I am not saying that to be diplomatic). With tests, I usually prefer to fail fast so you don't end up running tests that might be pointless. But with expensive end to end test, we are looking to extract as much information as possible from each run because they take time and resources.

Let's say we are running the test and somehow Bitbucket fails to start (e.g. license issue), but there is also an issue in our spec configuration for the NFS server. If we run only the _status endpoint test_, we will know only the first failure and we need to start an additional run to discover the NFS bug. If we run both tests, we should discover both issues in a single run. Because these tests are not tightly coupled (although there is some dependency between them) and they don't need to necessarily fail both at the same time, I think it is worth running both tests even if the first test fails because the setup is such an expensive operation.
-- my 2c :)

test/e2etest/installer_test.go
errcode1202: Added this line for each product test suite

PR #171, opened 2022-02-23
TFKUBE-344: Fix the re-installation by apply a unique subpath to shared home in installation

nghazali:


modules/products/bamboo/helm.tf
```diff
@@ -94,3 +94,9 @@ resource "helm_release" "bamboo_agent" {
     }),
   ]
 }
+
+resource "random_string" "random" {
+  length  = 10
```
errcode1202, line 99: Is ten a bit excessive? Maybe 5 is enough?

errcode1202 (approved):

modules/products/bamboo/helm.tf
nghazali: Yes, but still this is part of a path and is not visible at all (even for us) - the default is 16 and I made it 10.

PR #177, opened 2022-02-25
TFKUBE-333: Do not create EFS module if installing only Bitbucket

nghazali:


outputs.tf
```diff
@@ -22,7 +22,7 @@ output "efs" {
   description = "EFS shared storage"
 
   value = {
-    efs = module.base-infrastructure.efs.efs_id
+    efs = module.base-infrastructure.efs == null ? null : module.base-infrastructure.efs[0].efs_id
```
nanux, line 25:
```suggestion
    efs = module.base-infrastructure.efs == null ? "EFS is not provisioned" : module.base-infrastructure.efs[0].efs_id
```

locals.tf
```diff
@@ -7,6 +7,9 @@ locals {
   install_confluence = contains([for o in var.products : lower(o)], "confluence")
   install_bamboo     = contains([for o in var.products : lower(o)], "bamboo")
 
+  # If Bitbucket is the only product to install then we don't need to create shared home
+  share_home_size = local.install_bitbucket && length(var.products) == 1 ? "0Gi" : "5Gi"
```
nanux, line 11: I think the `0Gi` might be misleading.
 ```suggestion
  share_home_size = local.install_bitbucket && length(var.products) == 1 ? null : "5Gi"
```

Maybe even better to turn this into boolean:
```
should_create_efs = local.install_bitbucket && length(var.products) == 1 ? false : true
```

Comment (nghazali) [reply locals.tf:11]: I prefer to stick to `null` as we should fulfill the validation condition for `share_home_size` anyway.

File: modules/common/outputs.tf [hunk 3]
```diff
@@ -18,9 +18,8 @@ output "eks" {
   description = "EKS Module"
 }
 
-
 output "efs" {
-  value       = module.efs
+  value       = local.create_share_home && length(module.efs) == 1 ? module.efs : null
```
yzha645, line 22: Do we need to check `length(module.efs) == 1` while already checking `local.create_share_home`?

dc-infrastructure.tf
```diff
@@ -8,7 +8,7 @@ module "base-infrastructure" {
   desired_capacity = var.desired_capacity
   domain           = var.domain
   namespace        = local.namespace
-  share_home_size  = "5Gi"
+  share_home_size  = local.share_home_size
```
nanux, line 11: /nit I know it is not part of this PR but technically, we should be using `shared_home_size` (instead of `share_home_*` prefix.

nghazali: done

errcode1202 (changes requested):

modules/common/outputs.tf
nghazali: Actually is a good question. I have slightly changed this part and return `module.efs[0]` as `efs`, because now we have involved `count` in the module. In this case yes, we need to check the length to make sure there is no break in `plan` stage, when still the object is not created but terraform tried to reference it.

locals.tf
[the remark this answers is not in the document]
errcode1202: If this value is really for driving the size of shared home then it may be that the comment is what's throwing me.

modules/common/locals.tf
```diff
@@ -5,4 +5,5 @@ locals {
   ingress_domain = var.domain != null ? "${var.environment_name}.${var.domain}" : null
 
   storage_class_name = "efs-cs"
+  create_share_home  = !can(regex("^[0]([mMgG]|Gi)$", var.share_home_size))
```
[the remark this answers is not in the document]
nghazali: I changed this part and as the new implementation we check this with null.

locals.tf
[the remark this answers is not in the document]
nghazali: this is outdated code, please check the latest changes.

errcode1202 (approved):

modules/common/locals.tf
```diff
@@ -5,4 +5,5 @@ locals {
   ingress_domain = var.domain != null ? "${var.environment_name}.${var.domain}" : null
 
   storage_class_name = "efs-cs"
+  create_shared_home = var.shared_home_size != null
```
jjeongatl, line 8: It looks more straightforward if we create a new `create_shared_home` flag and pass from root level rather than checking shared_home_size.

jjeongatl (approved):

PR #178, opened 2022-02-28
TFKUBE-285: SSH for bitbucket

errcode1202:


modules/AWS/ingress/main.tf
```diff
@@ -71,6 +71,21 @@ resource "helm_release" "ingress" {
...
+      # https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/exposing-tcp-udp-services.md
+      #
+      # On deployment of the Ingress NGINX Helm chart the tcp stanza below will result in the creation of a config map,
+      # as describe above, including the addition of the --tcp-services-configmap flag to the controllers deployment.
+      # A port definition for 7999 will also be added to the controllers service.
+      #
+      # These 3 steps are effectively what is documented here:
+      # https://atlassian.github.io/data-center-helm-charts/examples/bitbucket/BITBUCKET_SSH/#nginx-ingress-controller-config-for-ssh-connections
+      tcp = {
+        7999: "atlassian/bitbucket:ssh"
```
errcode1202, line 87: This should only be added for Bitbucket deployments.

install.sh
```diff
@@ -260,6 +260,34 @@ set_synchrony_url() {
...
+  local load_balancer_name
+  local original_instance_port
+
+  install_bitbucket=$(get_product "bitbucket" "${CONFIG_ABS_PATH}")
+
+  if [ -n "${install_bitbucket}" ]; then
+    region=$(get_variable 'region' "${CONFIG_ABS_PATH}")
+    load_balancer_dns=$(terraform output | grep '"load_balancer_hostname" =' | sed -nE 's/^.*"(.*)".*$/\1/p')
+    load_balancer_name=$(echo "$load_balancer_dns" | cut -d '-' -f 1)
+    original_instance_port=$(aws elb describe-load-balancers --load-balancer-name "$load_balancer_name" --query 'LoadBalancerDescriptions[*].ListenerDescriptions[*].Listener[]' --region "$region" | jq '.[] | select(.LoadBalancerPort==7999) | .InstancePort')
```
errcode1202, line 276: Using the `query`/`filter` aws cli I wasn't successful in filtering out the listener with a `LoadBalancerPort` that matches `7999`. As a result Ive pulled in `jq`. I see we do have documentation that points to using `jq` so maybe this is okay? At the very least we would need to update the prerequisites to call out `jq` as a requirement.

errcode1202: Just also confirmed that `jq` is available to GitHub actions.

test/e2etest/bitbucket_test.go
```diff
@@ -46,3 +48,13 @@ func assertBitbucketNfsConnectivity(t *testing.T, testConfig TestConfig) {
 	assert.Nil(t, kubectlError)
 	assert.Equal(t, "Greetings from an NFS", fileContents)
 }
+
+func assertBitbucketSshConnectivity(t *testing.T, testConfig TestConfig) {
+	sshEndpoint := fmt.Sprintf("ssh://%s.%s.%s:7999", bitbucket, testConfig.EnvironmentName, domain)
+	println("Asserting Bitbucket SSH connectivity ...")
+	returnCode, sshError := exec.Command(
```
errcode1202, line 55: This test is still a work in progress...

install.sh
errcode1202: Added the function `check_for_prerequisites` to `install.sh` to check for prerequisites.

dc-infrastructure.tf
```diff
@@ -9,6 +9,8 @@ module "base-infrastructure" {
   domain           = var.domain
   namespace        = local.namespace
   share_home_size  = "5Gi"
+
+  enable_tcp = local.install_bitbucket
```
nanux, line 13:
```suggestion
  enable_ssh_tcp = local.install_bitbucket
```

I am assuming if there are some other products with TCP requirements they would have a separate flag. Also this means that the settings will need to be updated.

docs/docs/userguide/PREREQUISITES.md
```diff
@@ -9,8 +9,9 @@ Its advised that the tooling below is installed to your development environment.
 1. [Terraform](#terraform) 
 2. [Helm v3.3 or later](#helm)
 3. [AWS CLI](#aws-cli)
-4. [Kubectl](#kubectl) (optional)
-5. [Kubernetes cluster monitoring tools](#kubernetes-cluster-monitoring-tools) (optional)
+4. [jq ](#jq)
```
nanux, line 12:
```suggestion
4. [jq](#jq)
```

install.sh
```diff
@@ -260,6 +273,38 @@ set_synchrony_url() {
...
+  local load_balancer_name
+  local original_instance_port
+
+  install_bitbucket=$(get_product "bitbucket" "${CONFIG_ABS_PATH}")
+
+  if [ -n "${install_bitbucket}" ]; then
+    region=$(get_variable 'region' "${CONFIG_ABS_PATH}")
+    load_balancer_dns=$(terraform output | grep '"load_balancer_hostname" =' | sed -nE 's/^.*"(.*)".*$/\1/p')
+    load_balancer_name=$(echo "${load_balancer_dns}" | cut -d '-' -f 1)
+    original_instance_port=$(aws elb describe-load-balancers --load-balancer-name "${load_balancer_name}" --query 'LoadBalancerDescriptions[*].ListenerDescriptions[*].Listener[]' --region "${region}" | jq '.[] | select(.LoadBalancerPort==7999) | .InstancePort')
```
nanux, line 290: Today I have learned that the query language in `aws cli` is actually quite powerful https://jmespath.org/ I haven't used it much but with a bit of testing I was able to get to:

```suggestion
    original_instance_port=$(aws elb describe-load-balancers --load-balancer-name "${load_balancer_name}" --query 'LoadBalancerDescriptions[*].ListenerDescriptions[?Listener.LoadBalancerPort==`7999`].Listener[].InstancePort | [0]' --region "${region}"
```

This removes the dependency on `jq` completely and we can remove it from the documentation. I have tested this locally, but please verify it solve it with the changes as well.

docs/docs/userguide/PREREQUISITES.md
nanux: Seems that we don't need `jq`, see my comment in the script. We should be ok to use just `aws cli`.

docs/docs/userguide/PREREQUISITES.md
```diff
@@ -58,6 +59,17 @@ You need to have the AWS CLI tool installed on your local machine before creatin
     ```
 2. If the AWS CLI is not installed or you're running version 1, install AWS CLI version 2 by following the [official instructions](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html){.external}.
 
+### :material-code-json: jq
```
nanux, line 62: Remove see below.

install.sh
```diff
@@ -53,6 +53,19 @@ EOF
   shift $((${OPTIND} - 1))
   UNKNOWN_ARGS="$*"
 
+# Check for prerequisite tooling
+# https://atlassian-labs.github.io/data-center-terraform/userguide/PREREQUISITES/
+check_for_prerequisites() {
+  declare -a tools=("aws" "helm" "terraform" "jq")
```
nanux, line 59:
```suggestion
  declare -a tools=("aws" "helm" "terraform")
```

`jq` is not necessary

install.sh
errcode1202: So is jmspath built into the `aws cli`?

nanux: It is https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-filter.html

install.sh
errcode1202: Thanks @nanux for finding a solution to this - `jq` no longer needed: 
https://github.com/atlassian-labs/data-center-terraform/pull/178#r820328638

install.sh
```diff
@@ -260,6 +273,38 @@ set_synchrony_url() {
...
+  local load_balancer_name
+  local original_instance_port
+
+  install_bitbucket=$(get_product "bitbucket" "${CONFIG_ABS_PATH}")
+
+  if [ -n "${install_bitbucket}" ]; then
+    region=$(get_variable 'region' "${CONFIG_ABS_PATH}")
+    load_balancer_dns=$(terraform output | grep '"load_balancer_hostname" =' | sed -nE 's/^.*"(.*)".*$/\1/p')
+    load_balancer_name=$(echo "${load_balancer_dns}" | cut -d '-' -f 1)
+    original_instance_port=$(aws elb describe-load-balancers --load-balancer-name af4905e438c1d4ba2ada43c9eda1b879 --query 'LoadBalancerDescriptions[*].ListenerDescriptions[?Listener.LoadBalancerPort==`7999`].Listener[].InstancePort | [0]' --region "${region}")
```
nanux, line 290:
```suggestion
    original_instance_port=$(aws elb describe-load-balancers --load-balancer-name ${load_balancer_name} --query 'LoadBalancerDescriptions[*].ListenerDescriptions[?Listener.LoadBalancerPort==`7999`].Listener[].InstancePort | [0]' --region "${region}")
```

install.sh
```diff
@@ -260,6 +273,38 @@ set_synchrony_url() {
...
+
+    log "Enabling SSH connectivity for Bitbucket. Updating load balancer [${load_balancer_dns}] listener protocol from HTTP to TCP on port 7999..."
```
nanux, line 292: I think we can extract `7999` into some constant. There is a little chance it will change soon, but we are repeating this magic number several times.

nanux (approved):

PR #179, opened 2022-03-01
variable bamboo_license should be optional

jjeongatl:


errcode1202 (approved):

variables.tf
```diff
@@ -379,6 +379,7 @@ variable "bamboo_license" {
   description = "Bamboo license."
   type        = string
   sensitive   = true
+  default     = null
```
nanux, line 382: I believe we will need to change these resources as well:

https://github.com/atlassian-labs/data-center-terraform/blob/37823a0e386498dee0fc6f7165063b2cc5772c4b/modules/products/bamboo/locals.tf#L55

https://github.com/atlassian-labs/data-center-terraform/blob/073e98f445eb3d01880a008a63b530b1d30af638/modules/products/bamboo/secrets.tf#L19

nanux (changes requested):

variables.tf
nanux: And https://github.com/atlassian-labs/data-center-terraform/blob/61eb3eecb44d7acf9be5d2ddc8117b29850855d7/modules/products/bamboo/variables.tf#L44

errcode1202: On second thought, Bamboo needs to be fully pre-seeded so I dont believe this can be optional. What's the reason for the change @jjeongatl ?

errcode1202 (changes requested):
On second thought, Bamboo needs to be fully pre-seeded so I dont believe this can be optional. What's the reason for the change @jjeongatl ?

nghazali:
I applied some fix in installer itself. Bamboo should fully pre-seeded, however, the license should be optional just if the bamboo is not in the `products` list in config file. 

@jjeongatl , the fix you made is goos, just add a validation in Bamboo module for license variable to not accept `null` as value.

jjeongatl:
Closing this PR as the change will be in this [PR](https://github.com/atlassian-labs/data-center-terraform/pull/177)

PR #180, opened 2022-03-02
TFKUBE-349: End-to-end tests for Bitbucket

errcode1202:
Passing on branch plan: 
https://github.com/atlassian-labs/data-center-terraform/runs/5443327686?check_suite_focus=true


scripts/common.sh
```diff
@@ -58,4 +58,35 @@ get_product(){
...
+  else
+    echo "Usage: create_lb_listener function expects 3 params <load_balancer_name> <original_instance_port> <region>"
+  fi
+}
+
+describe_lb_listener() {
+  if [ "$#" -eq 2 ]; then
+    local load_balancer_name=${1}
+    local region=${2}
+    aws elb describe-load-balancers --load-balancer-name "$load_balancer_name" --query 'LoadBalancerDescriptions[*].ListenerDescriptions' --region "$region" | grep 7999 -B 2 -A 3
```
nghazali, line 88:
```suggestion
    local load_balancer_name="${1}"
    local region="${2}"
    aws elb describe-load-balancers --load-balancer-name "${load_balancer_name}" --query 'LoadBalancerDescriptions[*].ListenerDescriptions' --region "${region}" | grep 7999 -B 2 -A 3
```

scripts/common.sh
```diff
@@ -58,4 +58,35 @@ get_product(){
...
+    echo "Usage: delete_lb_listener function expects 2 params <load_balancer_name> <region>"
+  fi
+}
+
+create_lb_listener() {
+  if [ "$#" -eq 3 ]; then
+    local load_balancer_name=${1}
+    local original_instance_port=${2}
+    local region=${3}
+    aws elb create-load-balancer-listeners --load-balancer-name "$load_balancer_name" --listeners "Protocol=TCP,LoadBalancerPort=7999,InstanceProtocol=TCP,InstancePort=$original_instance_port" --region "$region"
```
nghazali, line 78:
```suggestion
    local load_balancer_name="${1}"
    local original_instance_port="${2}"
    local region="${3}"
    aws elb create-load-balancer-listeners --load-balancer-name "${load_balancer_name"} --listeners "Protocol=TCP,LoadBalancerPort=7999,InstanceProtocol=TCP,InstancePort=${original_instance_port}" --region "${region}"
```

scripts/common.sh
```diff
@@ -58,4 +58,35 @@ get_product(){
   fi
   echo "Usage: get_products <expected product> <config file>"
   return 1
+}
+
+delete_lb_listener() {
+  if [ "$#" -eq 2 ]; then
+    local load_balancer_name=${1}
+    local region=${2}
+    aws elb delete-load-balancer-listeners --load-balancer-name "$load_balancer_name" --load-balancer-ports 7999 --region "$region"
```
nghazali, line 67:
```suggestion
    local load_balancer_name="${1}"
    local region="${2}"
    aws elb delete-load-balancer-listeners --load-balancer-name "${load_balancer_name}" --load-balancer-ports 7999 --region "${region}"
```

.github/workflows/e2e-test.yaml
```diff
@@ -42,6 +42,18 @@ jobs:
...
+        uses: shimataro/ssh-key-action@v2
+        with:
+          key: ${{ secrets.BITBUCKET_E2E_TEST_PRIV_SSH_KEY }}
+          name: bitbucket-e2e # optional
+          known_hosts: dummy-entry
+          if_key_exists: fail # replace / ignore / fail; optional (defaults to fail)
+
+      - name: Add public SSH key for Bitbucket tests
+        run: |
+          echo ${{ secrets.BITBUCKET_E2E_TEST_PUB_SSH_KEY }} >> /home/runner/.ssh/bitbucket-e2e.pub
```
nghazali, line 55: Technically we should use `>` to create/overwrite the file for public key. For this case `>>` (append) works as well because this is in a Github action and we know no file `/home/runner/.ssh/bitbucket-e2e.pub` existed. 

However, I say as an optional change better to use the right syntax and replace `>>` with `>`

install.sh
```diff
@@ -260,6 +273,38 @@ set_synchrony_url() {
...
+  local region
+  local load_balancer_dns
+  local load_balancer_name
+  local original_instance_port
+
+  install_bitbucket=$(get_product "bitbucket" "${CONFIG_ABS_PATH}")
+
+  if [ -n "${install_bitbucket}" ]; then
+    region=$(get_variable 'region' "${CONFIG_ABS_PATH}")
+    load_balancer_dns=$(terraform output | grep '"load_balancer_hostname" =' | sed -nE 's/^.*"(.*)".*$/\1/p')
```
nghazali, line 288: I see we use `terraform output` in more than once in installer script (right before tagging process, when terraform apply is done). As running this command takes some time, and the result of multiple run is similar, maybe is good to run this command once and keep it in a variable in script and reuse it whenever is needed to improve the performance a bit.

install.sh
```diff
@@ -260,6 +273,38 @@ set_synchrony_url() {
...
+    load_balancer_name=$(echo "$load_balancer_dns" | cut -d '-' -f 1)
```
nghazali, line 289: Would you please change the address a variable in the script to the following pattern to be consistent:
```
"${variable}"
```

nghazali (changes requested):

install.sh
```diff
@@ -53,6 +53,19 @@ EOF
...
   UNKNOWN_ARGS="$*"
 
+# Check for prerequisite tooling
+# https://atlassian-labs.github.io/data-center-terraform/userguide/PREREQUISITES/
+check_for_prerequisites() {
+  declare -a tools=("aws" "helm" "terraform" "jq")
+  for tool in "${tools[@]}"
+  do :
+    if ! command -v "$tool" &>/dev/null; then
+      echo "The required dependency [$tool] could not be found. Please make sure that it is installed before continuing."
```
nghazali, line 63:
```suggestion
      echo "The required dependency [${tool}] could not be found. Please make sure that it is installed before continuing."
```

install.sh
nghazali: I couldn't find the code change for this comment. Or maybe we can optimise it as part of another PR. In this case good to raise a ticket to track this improvement.

errcode1202: 👍🏽  Created this: https://bulldog.internal.atlassian.com/browse/TFKUBE-368

yzha645 (approved):

nghazali (approved):

PR #181, opened 2022-03-07
TFKUBE-342: Make bamboo license optional when bamboo is not in products list

nghazali:


dc-infrastructure.tf
```diff
@@ -46,9 +46,10 @@ module "bamboo" {
     mem          = var.bamboo_mem
     min_heap     = var.bamboo_min_heap
     max_heap     = var.bamboo_max_heap
-    license      = var.bamboo_license
   }
 
+  license = var.bamboo_license
```
jjeongatl, line 51: Out of curiosity, What is the purpose of `bamboo_configuration` block?

jjeongatl (approved):

errcode1202 (approved):

nanux:
The changes LGTM. We need to fix unit tests before merging.

```
--- FAIL: TestBambooVariablesPopulatedWithValidValues (16.44s)
```

nanux (approved):

dc-infrastructure.tf
nghazali: This is for grouping the configuration settings. We had a discussion on this. Originally this is implemented to make the code more clear and organised. However, there is a point that we may better follow the common practise and use separated variables as other modules in market doing the same. However, we don't expose internal modules (module folder and products) this method does not violate the common practise but still this is an option for refactor in future.

PR #182, opened 2022-03-07
TFKUBE-201: Add cluster-autoscaler

nanux:
* add cluster-autoscaler helm release
* create IAM role with necessary permission policy to scale the EKS cluster
* use single subnet for the EKS node group
* enable IRSA (IAM Roles for Service Accounts) in the EKS cluster
* upgrade IAM module to 4.13.2
* add region variable to the EKS module
* add helm provide for EKS module
* add unit tests
* add e2e test case for autoscaler
* decrease the initial e2e capacity to 1 (and expect autoscaler to increase the number of nodes)
* fix developer docs

### DIFF
--- a/modules/AWS/eks/autoscaling.tf
+++ b/modules/AWS/eks/autoscaling.tf
@@ -1,0 +1,102 @@
+module "autoscaler_iam_role" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version = "4.13.2"
+
+  create_role  = true
+  role_name    = "${var.cluster_name}-cluster-autoscaler"
+  provider_url = replace(module.eks.cluster_oidc_issuer_url, "https://", "")
+
+  role_policy_arns = [
+    aws_iam_policy.cluster_autoscaler.arn
+  ]
+  oidc_fully_qualified_subjects = [
+    "system:serviceaccount:${local.autoscaler_service_account_namespace}:${local.autoscaler_service_account_name}"
+  ]
+}
+
+resource "aws_iam_policy" "cluster_autoscaler" {
+  name_prefix = "cluster-autoscaler"
+  description = "EKS cluster-autoscaler policy for cluster ${module.eks.cluster_id}"
+  policy      = data.aws_iam_policy_document.cluster_autoscaler.json
+}
+
+data "aws_iam_policy_document" "cluster_autoscaler" {
+  statement {
+    sid    = "clusterAutoscalerAll"
+    effect = "Allow"
+
+    actions = [
+      "autoscaling:DescribeAutoScalingGroups",
+      "autoscaling:DescribeAutoScalingInstances",
+      "autoscaling:DescribeLaunchConfigurations",
+      "autoscaling:DescribeTags",
+      "ec2:DescribeLaunchTemplateVersions",
+      "ec2:DescribeInstanceTypes"
+    ]
+
+    resources = [
+      "*"
+    ]
+  }
+
+  statement {
+    sid    = "clusterAutoscalerOwn"
+    effect = "Allow"
+
+    actions = [
+      "autoscaling:SetDesiredCapacity",
+      "autoscaling:TerminateInstanceInAutoScalingGroup",
+      "autoscaling:UpdateAutoScalingGroup",
+    ]
+
+    resources = [
+      "*"
+    ]
+
+    condition {
+      test     = "StringEquals"
+      variable = "autoscaling:ResourceTag/kubernetes.io/cluster/${module.eks.cluster_id}"
+      values = [
+        "owned"
+      ]
+    }
+
+    condition {
+      test     = "StringEquals"
+      variable = "autoscaling:ResourceTag/k8s.io/cluster-autoscaler/enabled"
+      values = [
+        "true"
+      ]
+    }
+  }
+}
+
+resource "helm_release" "cluster-autoscaler" {
+  depends_on = [
+    module.eks
+  ]
+
+  name       = "cluster-autoscaler"
+  namespace  = local.autoscaler_service_account_namespace
+  repository = "https://kubernetes.github.io/autoscaler"
+  chart      = "cluster-autoscaler"
+  version    = local.autoscaler_version
+
+  values = [yamlencode({
+    awsRegion = var.region
+    rbac = {
+      create = true
+      serviceAccount = {
+        name = local.autoscaler_service_account_name
+        annotations = {
+          "eks.amazonaws.com/role-arn" : module.autoscaler_iam_role.iam_role_arn
+        }
+      }
+    }
+    autoDiscovery = {
+      enabled     = true
+      clusterName = var.cluster_name
+    }
+  })]
+}
+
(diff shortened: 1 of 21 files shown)


nanux:
I have kicked off an end to end test that starts with 1 EC2 instance and left it up to autoscaler to add more nodes. https://github.com/atlassian-labs/data-center-terraform/actions/runs/1943941571

modules/AWS/eks/autoscaling.tf
```diff
@@ -0,0 +1,102 @@
+module "autoscaler_iam_role" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version = "4.13.2"
+
+  create_role  = true
+  role_name    = "${var.cluster_name}-cluster-autoscaler"
```
nghazali, line 6: cluster name has `-cluster` as postfix so we don't have to repeat it here. 
```suggestion
  role_name    = "${var.cluster_name}-autoscaler"
```

modules/AWS/eks/autoscaling.tf
```diff
@@ -0,0 +1,102 @@
...
+    rbac = {
+      create = true
+      serviceAccount = {
+        name = local.autoscaler_service_account_name
+        annotations = {
+          "eks.amazonaws.com/role-arn" : module.autoscaler_iam_role.iam_role_arn
+        }
+      }
+    }
+    autoDiscovery = {
```
nghazali, line 96: Shouldn't we define `autoscalingGroups` instead of `autoDiscovery` to force limit for scaling?

test/e2etest/test-config.tfvars.tmpl
```diff
@@ -29,7 +29,7 @@ resource_tags = {
 instance_types = ["m5.xlarge"]
 
 # Desired number of nodes that the node group should launch with initially.
-desired_capacity = 3
+desired_capacity = 1
```
nghazali, line 32: We did this on purpose to test multiple EKS nodes in e2e test. Back desired capacity to one will result failure in e2e test as the instance type has not enough resources to provide to all products. If you change this to `1` you need to change the instance type that I don't suggest it. I think we better revert this change and leave it as `3`.

modules/AWS/eks/autoscaling.tf
nanux: Why do you think this is a problem? I did follow the official documentation which is using autodiscovery:

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup

> Auto-Discovery Setup is the preferred method to configure Cluster Autoscaler.

I am happy to change if it will be beneficial but I am not sure if we gain some improvement. The current setup also adheres to the maximum capacity of the used ASG. I have tested it by adding another node. Note the `3->4 (max: 10)` message:

```
I0308 06:19:48.682635       1 scale_up.go:468] Best option to resize: eks-appNode-m5_2xlarge-2abfb126-211a-9c36-aa0d-a9c663e33a07
I0308 06:19:48.682648       1 scale_up.go:472] Estimated 1 nodes needed in eks-appNode-m5_2xlarge-2abfb126-211a-9c36-aa0d-a9c663e33a07
I0308 06:19:48.682668       1 scale_up.go:595] Final scale-up plan: [{eks-appNode-m5_2xlarge-2abfb126-211a-9c36-aa0d-a9c663e33a07 3->4 (max: 10)}]
I0308 06:19:48.682677       1 scale_up.go:691] Scale-up: setting group eks-appNode-m5_2xlarge-2abfb126-211a-9c36-aa0d-a9c663e33a07 size to 4
```

test/e2etest/test-config.tfvars.tmpl
nanux: This is a part of the end to end test. We will start with the smaller capacity and expect the autoscaler to spin up additional resources. The failures we were having were caused by the missing IAM permissions but after fixing those, we have now a passing build:

https://github.com/atlassian-labs/data-center-terraform/actions/runs/1943941571

We have also added an explicit test for the cluster autoscaler that is more defining our expectations:
https://github.com/atlassian-labs/data-center-terraform/pull/182/files#diff-6afb0bd2b96fbd68692fc79175b6bc446cb66721bbe7f07fb693df9d57fc5bdfR10

The e2e build is now running:
https://github.com/atlassian-labs/data-center-terraform/actions/runs/1949842089

modules/AWS/eks/autoscaling.tf
nghazali: make sense

nghazali (approved):

nghazali:
Looks good to me, just we should make sure e2e test is passed after this changes before merge the PR.

nanux:
> Looks good to me, just we should make sure e2e test is passed after this changes before merge the PR.

There has been another [successful run](https://github.com/atlassian-labs/data-center-terraform/actions/runs/1956532442) and I have kicked off another one to make sure it is not flaky.

nanux:
The workflow was successful. https://github.com/atlassian-labs/data-center-terraform/actions/runs/1956836096

I think this is ready to be fully reviewed and merged.

providers.tf
```diff
@@ -1,3 +1,17 @@
+terraform {
```
nanux, line 1: This pins the versions for the main providers (e2e test is passing).
https://github.com/atlassian-labs/data-center-terraform/runs/5478975078?check_suite_focus=true#step:8:103

errcode1202 (approved):

yzha645 (approved):

nghazali:
After this PR is merged we need to update the documentation accordingly. Maybe we should update the architect diagram in README.md, and also Currently we have some context in `Common configuration` page that suggests EKS nodes count cannot be changed, and so which needs to be updated.

modules/AWS/eks/main.tf
```diff
@@ -1,20 +1,22 @@
...
   source  = "terraform-aws-modules/eks/aws"
-  version = "17.23.0"
+  version = "17.24.0"
 
   # Configure cluster
   cluster_version              = "1.21"
   cluster_name                 = var.cluster_name
   manage_cluster_iam_resources = true
 
+  enable_irsa = true
```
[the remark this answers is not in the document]
nanux: 🪄

modules/AWS/eks/main.tf
```diff
@@ -25,11 +27,12 @@ module "eks" {
 
   node_groups = {
     appNodes = {
-      name             = "appNode"
+      name             = "appNode-${replace(join("-", var.instance_types), ".", "_")}"
```
jjeongatl, line 30: Is this to make it more descriptive or required for auto scaling?

providers.tf
jjeongatl: We should put aws pin on modules/tfstate/main.tf

nanux: Probably yes because that is a separate module.

modules/AWS/eks/main.tf
nanux: This is to make it more descriptive but also enable the change of the instance_type. I haven't tested that thoroughly but it will get rid of the static `appNode` string. When you change the instance type, a new node group is created and after it is created, the workload is migrated there and the previous node group is destroyed.

PR #183, opened 2022-03-08
TFKUBE-330: Bitbucket display name

yzha645:
This PR added support for Bitbucket display name.

### DIFF
--- a/config.tfvars
+++ b/config.tfvars
@@ -130,6 +130,9 @@ bitbucket_helm_chart_version = "1.2.0"
 #bitbucket_admin_display_name  = "<DISPLAY NAME>"
 #bitbucket_admin_email_address = "<EMAIL ADDRESS>"
 
+# The display name of Bitbucket instance
+#bitbucket_display_name = "<DISPLAY NAME>"
+
 # Bitbucket instance resource configuration
 bitbucket_cpu      = "1"
 bitbucket_mem      = "1Gi"
--- a/config.tfvars
+++ b/config.tfvars
@@ -79,7 +79,7 @@ confluence_helm_chart_version = "1.2.0"
 # To avoid storing license in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_` (i.e. `TF_VAR_confluence_license`) and keep the below line commented out
 # If storing license as plain-text is not a concern for this environment, feel free to uncomment the following line and supply the license here
 #
-#confluence_license = "<LICENSE KEY>"
+#confluence_license = "<LICENSE_KEY>"
 
 # Confluence instance resource configuration
 confluence_cpu      = "1"
@@ -116,7 +116,7 @@ bitbucket_helm_chart_version = "1.2.0"
 # To avoid storing license in a plain text file, we recommend storing it in an environment variable prefixed with `TF_VAR_` (i.e. `TF_VAR_bitbucket_license`) and keep the below line commented out
 # If storing license as plain-text is not a concern for this environment, feel free to uncomment the following line and supply the license here
 #
-#bitbucket_license = "<LICENSE KEY>"
+#bitbucket_license = "<LICENSE_KEY>"
 
 # Bitbucket system admin credentials
 # To pre-seed Bitbucket with the system admin information, uncomment the following settings and supply the system admin information:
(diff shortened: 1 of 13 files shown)


config.tfvars
```diff
@@ -130,6 +130,9 @@ bitbucket_helm_chart_version = "1.2.0"
 #bitbucket_admin_display_name  = "<DISPLAY NAME>"
 #bitbucket_admin_email_address = "<EMAIL ADDRESS>"
 
+# The display name of Bitbucket instance
+#bitbucket_display_name = "<DISPLAY NAME>"
```
errcode1202, line 134: Need a space between `#` and comment

docs/docs/userguide/configuration/BITBUCKET_CONFIGURATION.md
```diff
@@ -50,6 +49,13 @@ bitbucket_admin_email_address = "<EMAIL ADDRESS>"
 
     Please refer to [Sensitive Data](#sensitive-data) section.
 
+### Display Name
+Set the display name of the Bitbucket instance. Note that this value is only used during installation and changing the value during an upgrade has no effect.
```
errcode1202, line 53: Out of interest what does setting the display name actually do - is it a UI feature? Maybe the comment should describe its purpose.

docs/docs/userguide/configuration/BITBUCKET_CONFIGURATION.md
```diff
@@ -50,6 +49,13 @@ bitbucket_admin_email_address = "<EMAIL ADDRESS>"
...
+
+```terraform
+bitbucket_display_name = "<DISPLAY NAME>"
```
errcode1202, line 56: I think the current pattern is to use `_`  to delineate instead of space.

modules/products/bitbucket/locals.tf
```diff
@@ -77,4 +77,11 @@ locals {
       { name = "cluster.initial_master_nodes", value = "" }
     ]
   })
+
+  # Bitbucket display  name
```
errcode1202, line 81: extra space between `display  name`

errcode1202 (approved):

config.tfvars
yzha645: If user uses `command`+`/` to uncomment this line, an extra space will make it unaligned... 
And to be consistent with other commented-out variables, see line 130, 131.

docs/docs/userguide/configuration/BITBUCKET_CONFIGURATION.md
yzha645: yea.. Seems like we have inconsistent patterns across the doc. [Example 1](https://github.com/atlassian-labs/data-center-terraform/blob/main/docs/docs/userguide/configuration/CONFIGURATION.md#domain), [example 2](https://github.com/atlassian-labs/data-center-terraform/blob/main/docs/docs/userguide/configuration/BITBUCKET_CONFIGURATION.md#database-instance-class)
For this specific one, I kinda like the space delimiter.. since the value can actually have spaces in between, e.g. `Yifei's Stage Instance`.

docs/docs/userguide/configuration/BITBUCKET_CONFIGURATION.md
yzha645: yup, will elaborate it more. And display name will be showing up in the browser tab: 
Screen Shot 2022-03-09 at 2 26 57 pm

nanux: This is the description in the documentation (but I don't think it is great). https://confluence.atlassian.com/bitbucketserver/automated-setup-for-bitbucket-776640098.html

```
The display name for the Bitbucket application.
```

I think providing more detail can be useful.

nanux (approved):

docs/docs/userguide/configuration/BITBUCKET_CONFIGURATION.md
yzha645: I tried to find some official document on `display name`, but only found this one, very similar to what we got now: 
https://confluence.atlassian.com/bitbucketserver/automated-setup-for-bitbucket-776640098.html 

I'm thinking while we're not sure what other places will use `display name`, will it be misleading to try defining it here?

modules/products/bitbucket/variables.tf
```diff
@@ -128,3 +128,8 @@ variable "elasticsearch_replicas" {
   }
 }
 
+variable "display_name" {
+  description = "The display name of Bitbucket instance."
+  type        = string
+  default     = null
+}
```
nghazali, line 135: Do we have any restriction for display name, such as length or any special character? In this case it would be good to add validation here or in root level (preferably).

nghazali (approved):

nghazali:
Just left a minor comment. Other than this all are already discussed. Thank you

modules/products/bitbucket/variables.tf
yzha645: That's actually a good point. I rechecked Bitbucket Helm chart, there is no length restriction on this field.. So I guess for now, the `string` type restriction should be enough?

PR #184, opened 2022-03-09
TFKUBE-375: Add replica count

nanux:
* Add replica_count variables to product modules
* Add product_replica_count to config
* Fix unit tests

### Test case
1. Installed all products
2. Bitbucket had fully automated setup with 3 replicas, the app starts with 3 nodes.
3. Jira and Confluence were installed with 1 replica and finished the setup in UI
4. Scaled up the replicaCount for Jira and Confluence to 3

### Results
The autoscaler kicked in and spun up a new node. The pods were added to the statefulSet and joined the cluster.

Screen Shot 2022-03-09 at 10 15 56 pm
Screen Shot 2022-03-09 at 10 15 50 pm
Screen Shot 2022-03-09 at 10 15 42 pm

**Timing:**
```
module.confluence[0].helm_release.confluence: Modifications complete after 4m16s [id=confluence]
module.jira[0].helm_release.jira: Modifications complete after 4m39s [id=jira]
```


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
nghazali, line 1: Should we include this file? This is a IDE setting file.

variables.tf
```diff
@@ -91,6 +91,12 @@ variable "jira_version_tag" {
   default     = null
 }
 
+variable "jira_replica_count" {
+  description = "Number of Jira application nodes"
+  type        = number
+  default     = 1
+}
```
nghazali, line 98: It would be nice to add a validator for a range of `[0-max-pods#]`

nghazali (approved):

nghazali:
Left some minor comments. Other than that it looks good to go.

variables.tf
nanux: Good point! There is no upper limit and you can also scale the `replicaCount` to zero, so I have added only `>= 0` validation.

errcode1202 (approved):

PR #185, opened 2022-03-10
TFKUBE-330: Bitbucket display name validation

yzha645:
Added validation for Bitbucket display name to be limited to less than 255 characters.

### DIFF
--- a/modules/products/bitbucket/variables.tf
+++ b/modules/products/bitbucket/variables.tf
@@ -47,6 +47,11 @@ variable "db_iops" {
   type        = number
 }
 
+variable "replica_count" {
+  description = "Number of Bitbucket application nodes"
+  type        = number
+}
+
 variable "bitbucket_configuration" {
   description = "Bitbucket resource spec and chart version"
   type        = map(any)
--- a/modules/products/bitbucket/variables.tf
+++ b/modules/products/bitbucket/variables.tf
@@ -132,4 +132,8 @@ variable "display_name" {
   description = "The display name of Bitbucket instance."
   type        = string
   default     = null
+  validation {
+    condition     = length(var.display_name) > 0 && length(var.display_name) < 255
+    error_message = "Bitbucket display name must be a non-empty value less than 255 characters."
+  }
 }
--- a/modules/products/bitbucket/variables.tf
+++ b/modules/products/bitbucket/variables.tf
@@ -133,7 +133,7 @@ variable "display_name" {
   type        = string
   default     = null
   validation {
-    condition     = var.display_name == null || can(length(var.display_name) > 0 && length(var.display_name) < 255)
+    condition     = tobool(length(var.display_name) > 0 && length(var.display_name) < 255) || var.display_name == null
     error_message = "Bitbucket display name must be a non-empty value less than 255 characters."
   }
 }
--- a/modules/products/bitbucket/variables.tf
+++ b/modules/products/bitbucket/variables.tf
@@ -131,9 +131,9 @@ variable "elasticsearch_replicas" {
 variable "display_name" {
   description = "The display name of Bitbucket instance."
   type        = string
-  default     = null
+  default     = ""
   validation {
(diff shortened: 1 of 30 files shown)


modules/products/bitbucket/variables.tf
```diff
@@ -132,4 +132,8 @@ variable "display_name" {
   description = "The display name of Bitbucket instance."
   type        = string
   default     = null
+  validation {
+    condition     = length(var.display_name) > 0 && length(var.display_name) < 255
```
nghazali, line 136: I think display name is not mandatory field could be null (we have null as default). Please add null as an accepted condition in the validation and update the error message.

variables.tf
```diff
@@ -304,6 +304,10 @@ variable "bitbucket_display_name" {
   description = "The display name of Bitbucket instance"
   type        = string
   default     = null
+  validation {
```
nghazali, line 307: One validation is sufficient. Because we not suggesting to use the modules individually, having one validation in root level would be sufficient and visible. The maintenance would be easier if we need to change the validations in future as well.

errcode1202: Terraform question - if this validation is at this level does it mean invalid strings caught and flagged sooner?

errcode1202 (approved):

modules/products/bitbucket/variables.tf
yzha645: Good spot!

variables.tf
nghazali: not really, because validation will happening in planning stage. And terraform will recursively look for all modules to complete the plan. But is good to put the validations in root level to be more visible to users. They just need to look at the root variables.tf to see the validations, and they will not hide in module structures. 
Having the validation in module level is very good when modules can be re-use individually, but in our project, we do not intend to produce the modules to be used as a single unit.

yzha645: yup, fair enough~ But instead of root module, I'd like to keep it in Bitbucket module.

nghazali: @yzha645 , following our chat, you pointed a good point that we use the validation in unit test, so it would be fine to have it in module level, but please remove the redundant validation code from root level. Repeating the same code in two files makes us a difficult life in maintaining the code.

modules/products/bitbucket/variables.tf
nghazali: One more thing, check `length(var.display_name) < 255`  is sufficient check here, we don't need to check if the length is more than 0 as empty string is still a valid input here.

nanux: I don't think empty string is valid here:
Screen Shot 2022-03-10 at 4 29 18 pm

nanux (approved):

nghazali (changes requested):

nghazali (approved):

modules/products/bitbucket/variables.tf
```diff
@@ -136,5 +136,9 @@ variable "elasticsearch_replicas" {
 variable "display_name" {
   description = "The display name of Bitbucket instance."
   type        = string
-  default     = null
+  default     = ""
+  validation {
+    condition     = length(var.display_name) >= 0 && length(var.display_name) < 255
```
nghazali, line 141: This will work with `null` as default:

```
default     = ""
  validation {
    condition     = var.display_name == null || can(regex("^([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$", var.display_name))
    error_message = "Bitbucket display name must be a non-empty value less than 255 characters."
  }
```

yzha645: You're absolutely right! @nghazali 
`can()` and `regex()` combo is a powerful tool, given that regex() throws error when not matching. 
https://www.terraform.io/language/functions/can 
https://www.terraform.io/language/functions/regex 

I've modified the code, if you're happy about the change, do you mind helping merge this PR on Monday?

nghazali: @yzha645 , the modification you made in regx is brilliant. Well done! 🚀

PR #193, opened 2022-03-31
TFKUBE-401:  Jira snapshot restore

nghazali:
## Pull request description

## Checklist

### DIFF
--- a/config.tfvars
+++ b/config.tfvars
@@ -77,15 +77,17 @@ jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
 
-# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier. This will
-# restore the database from the snapshot and will not create a new database.
+# Dataset restore configuration
+# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier.
+# This will restore the database from the snapshot and will not create a new database.
+# The snapshot should be in the same AWS region as the environment and be accessible with the your AWS account.
 #jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"
 
-# (Optional) The master username for the database instance. If you provided, the username is "postgres".
-#jira_rds_master_username     = "<RDS_MASTER_USERNAME>"
-
-# (Optional) The master password for the database instance. If not provided, a random password will be generated.
-#jira_rds_master_password     = "<RDS_MASTER_PASSWORD>"
+# (Optional) The master user credential for the database instance.
+# If username not provided, the username is "postgres".
+# If password not provided, a random password will be generated.
+#jira_db_master_username     = "<DB_MASTER_USERNAME>"
+#jira_db_master_password     = "<DB_MASTER_PASSWORD>"
 
 ################################################################################
 # Confluence Settings
--- a/config.tfvars
+++ b/config.tfvars
@@ -81,8 +81,11 @@ jira_db_iops                 = 1000
 # restore the database from the snapshot and will not create a new database.
 #jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER"
 
+# (Optional) The master username for the database instance. If you provided, the username is "postgres".
+#jira_rds_master_username     = "<RDS_MASTER_USERNAME>!"
+
 # (Optional) The master password for the database instance. If not provided, a random password will be generated.
-#jira_db_master_password     = "<DB_MASTER_PASSWORD>!"
+#jira_rds_master_password     = "<RDS_MASTER_PASSWORD>!"
 
 ################################################################################
 # Confluence Settings
--- a/config.tfvars
+++ b/config.tfvars
@@ -82,10 +82,10 @@ jira_db_iops                 = 1000
 #jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"
 
 # (Optional) The master username for the database instance. If you provided, the username is "postgres".
-#jira_rds_master_username     = "<RDS_MASTER_USERNAME>!"
+#jira_rds_master_username     = "<RDS_MASTER_USERNAME>"
 
 # (Optional) The master password for the database instance. If not provided, a random password will be generated.
-#jira_rds_master_password     = "<RDS_MASTER_PASSWORD>!"
+#jira_rds_master_password     = "<RDS_MASTER_PASSWORD>"
 
 ################################################################################
 # Confluence Settings
--- a/config.tfvars
+++ b/config.tfvars
@@ -77,6 +77,13 @@ jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
 
+# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier. This will
+# restore the database from the snapshot and will not create a new database.
+#jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER"
+
+# (Optional) The master password for the database instance. If not provided, a random password will be generated.
+#jira_db_master_password     = "<DB_MASTER_PASSWORD>!"
+
 ################################################################################
 # Confluence Settings
 ################################################################################
--- a/dc-infrastructure.tf
+++ b/dc-infrastructure.tf
@@ -84,8 +84,8 @@ module "jira" {
   db_instance_class       = var.jira_db_instance_class
   db_iops                 = var.jira_db_iops
   db_snapshot_identifier  = var.jira_db_snapshot_identifier
-  db_master_username      = var.jira_rds_master_username
-  db_master_password      = var.jira_rds_master_password
+  db_master_username      = var.jira_db_master_username
+  db_master_password      = var.jira_db_master_password
 
   pvc_claim_name = module.base-infrastructure.pvc_claim_name
 
--- a/dc-infrastructure.tf
+++ b/dc-infrastructure.tf
@@ -83,6 +83,7 @@ module "jira" {
   db_allocated_storage    = var.jira_db_allocated_storage
   db_instance_class       = var.jira_db_instance_class
   db_iops                 = var.jira_db_iops
+  db_snapshot_identifier  = var.jira_db_snapshot_identifier
 
   pvc_claim_name = module.base-infrastructure.pvc_claim_name
 
--- a/dc-infrastructure.tf
+++ b/dc-infrastructure.tf
@@ -84,6 +84,7 @@ module "jira" {
   db_instance_class       = var.jira_db_instance_class
   db_iops                 = var.jira_db_iops
   db_snapshot_identifier  = var.jira_db_snapshot_identifier
+  db_master_pwd           = var.jira_db_master_pwd
 
   pvc_claim_name = module.base-infrastructure.pvc_claim_name
 
--- a/dc-infrastructure.tf
+++ b/dc-infrastructure.tf
@@ -84,7 +84,8 @@ module "jira" {
   db_instance_class       = var.jira_db_instance_class
   db_iops                 = var.jira_db_iops
   db_snapshot_identifier  = var.jira_db_snapshot_identifier
-  db_master_password      = var.jira_db_master_password
+  db_master_username      = var.jira_rds_master_username
+  db_master_password      = var.jira_rds_master_password
 
   pvc_claim_name = module.base-infrastructure.pvc_claim_name
 
--- a/dc-infrastructure.tf
+++ b/dc-infrastructure.tf
@@ -84,7 +84,7 @@ module "jira" {
   db_instance_class       = var.jira_db_instance_class
   db_iops                 = var.jira_db_iops
   db_snapshot_identifier  = var.jira_db_snapshot_identifier
-  db_master_pwd           = var.jira_db_master_pwd
+  db_master_password      = var.jira_db_master_password
 
   pvc_claim_name = module.base-infrastructure.pvc_claim_name
 
--- a/docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
+++ b/docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
@@ -99,3 +99,11 @@ jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"   # e.g. "my-snapshot"
 ```
 
 !!! info "The DB snapshot must be in the same AWS Region and AWS account as the DB instance. If you are restoring from a shared DB snapshot, the DB instance must have backup retention enabled."
+
+### Database Master Password
+
+'jira_db_master_password' sets the password for the RDS master user. If you do not specify a value, a random password will be generated.
+
+```terraform
+jira_db_master_password = "<DB_MASTER_PASSWORD>"   # default value is null
+```
--- a/docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
+++ b/docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
@@ -89,3 +89,13 @@ jira_db_iops = 1000
 
 !!! info "The allowed value range of IOPS may vary based on instance class"
 You may want to adjust these values according to your needs. For more information, see [Amazon RDS DB instance storage — Amazon Relational Database Service](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html){.external}.
+
+### Database Snapshot Identifier
+
+`jira_db_snapshot_identifier` sets the identifier for the DB snapshot to restore from. If you do not specify a value, no DB snapshot is used.
+
+```terraform
+jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"   # e.g. "my-snapshot"
+```
+
+!!! info "The DB snapshot must be in the same AWS Region and AWS account as the DB instance. If you are restoring from a shared DB snapshot, the DB instance must have backup retention enabled."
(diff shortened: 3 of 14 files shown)


errcode1202 (commented):
Im wondering if, lifecycle management and index triggering should be included in this PR or separate ones?

dc-infrastructure.tf
```diff
@@ -83,6 +83,8 @@ module "jira" {
   db_allocated_storage    = var.jira_db_allocated_storage
   db_instance_class       = var.jira_db_instance_class
   db_iops                 = var.jira_db_iops
+  db_snapshot_identifier  = var.jira_db_snapshot_identifier
+  db_master_pwd           = var.jira_db_master_pwd
```
jjeongatl, line 87: renaming it to `db_master_password` will keep the consistency

jjeongatl (changes requested):

errcode1202 (changes requested):

modules/AWS/rds/variables.tf
```diff
@@ -45,3 +45,15 @@ variable "major_engine_version" {
...
 }
+
+variable "snapshot_identifier" {
+  description = "Snapshot identifier for RDS."
+  type        = string
+  default     = null
+}
+
+variable "db_master_pwd" {
+  description = "Master password for the RDS instance."
```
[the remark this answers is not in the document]
nghazali: Here in RDS module we add master password as a generic option. In Jira or any other product that can pass snapshot db they can define a product password variable for RDS and pass it to RDS module when they create the database in product level.

modules/AWS/rds/variables.tf
```diff
@@ -45,3 +45,15 @@ variable "major_engine_version" {
     error_message = "Invalid major engine version. Valid ranges are from 10 to 13 (integer)."
   }
 }
+
+variable "snapshot_identifier" {
+  description = "Snapshot identifier for RDS."
```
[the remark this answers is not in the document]
nghazali: I add more detail here.

variables.tf
```diff
@@ -177,6 +177,22 @@ variable "jira_db_iops" {
   type        = number
 }
 
+variable "jira_db_snapshot_identifier" {
```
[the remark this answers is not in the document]
nghazali: I believe we shouldn't have to expose it directly in `config.tfvars`. 

The use case for using snapshot is just for running toolkits and `dc-app` team will create a config file and provides to the vendors to use. Add this in `config.tfvars` may bring some confusion and unnecessary complexity.

dc-infrastructure.tf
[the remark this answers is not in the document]
nghazali: Is already applied

config.tfvars
```diff
@@ -77,6 +77,13 @@ jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
 
+# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier. This will
+# restore the database from the snapshot and will not create a new database.
+#jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER"
```
jjeongatl, line 82:
```suggestion
#jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"
```

config.tfvars
```diff
@@ -77,6 +77,13 @@ jira_db_instance_class       = "db.t3.micro"
...
+
+# (Optional) The master password for the database instance. If not provided, a random password will be generated.
+#jira_db_master_password     = "<DB_MASTER_PASSWORD>!"
```
jjeongatl, line 85:
```suggestion
#jira_db_master_password     = "<DB_MASTER_PASSWORD>"
```

variables.tf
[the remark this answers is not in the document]
nghazali: Done

jjeongatl:
Should we add db snapshot restoration to e2e test?

modules/AWS/rds/locals.tf
```diff
@@ -1,5 +1,6 @@
 locals {
-  db_master_usr      = "${var.product}user"
+  db_master_usr      = "atl${var.product}"
+  db_master_pwd      = var.db_master_pwd == null ? random_password.password.result : var.db_master_pwd
```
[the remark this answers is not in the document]
nghazali: This is actually master user for RDS instance.

modules/AWS/rds/variables.tf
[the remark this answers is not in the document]
errcode1202: Makes sense!

config.tfvars
```diff
@@ -77,6 +77,16 @@ jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
 
+# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier. This will
+# restore the database from the snapshot and will not create a new database.
+#jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"
+
+# (Optional) The master username for the database instance. If you provided, the username is "postgres".
+#jira_rds_master_username     = "<RDS_MASTER_USERNAME>!"
```
errcode1202, line 85: Is the `!` at the end valid?

errcode1202 (changes requested):

config.tfvars
jjeongatl, line 85:
```suggestion
#jira_rds_master_username     = "<RDS_MASTER_USERNAME>"
```

config.tfvars
```diff
@@ -77,6 +77,16 @@ jira_db_instance_class       = "db.t3.micro"
...
+
+# (Optional) The master password for the database instance. If not provided, a random password will be generated.
+#jira_rds_master_password     = "<RDS_MASTER_PASSWORD>!"
```
jjeongatl, line 88:
```suggestion
#jira_rds_master_password     = "<RDS_MASTER_PASSWORD>"
```

variables.tf
```diff
@@ -177,6 +177,32 @@ variable "jira_db_iops" {
...
+  default     = null
+  type        = string
+}
+
+variable "jira_rds_master_username" {
+  description = "Master username for the Jira RDS instance."
+  type        = string
+  default     = null
+  validation {
+    condition     = can(regex("^[a_zA-Z_]\\w.{5,30}$", var.jira_rds_master_username)) || var.jira_rds_master_username == null
```
jjeongatl, line 191: `postgres` didn't pass the validation

jjeongatl, line 191:
```suggestion
    condition     = can(regex("^[a-zA-Z_].{5,30}$", var.jira_rds_master_username)) || var.jira_rds_master_username == null
```

jjeongatl (changes requested):

errcode1202 (changes requested):

variables.tf
nghazali: fixed

errcode1202 (approved):

jjeongatl (approved):

config.tfvars
```diff
@@ -77,6 +77,16 @@ jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
 
+# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier. This will
+# restore the database from the snapshot and will not create a new database.
+#jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"
+
+# (Optional) The master username for the database instance. If you provided, the username is "postgres".
+#jira_rds_master_username     = "<RDS_MASTER_USERNAME>"
```
nanux, line 85:
```suggestion
#jira_ds_master_username     = "<DB_MASTER_USERNAME>"
```

config.tfvars
```diff
@@ -77,6 +77,16 @@ jira_db_instance_class       = "db.t3.micro"
...
+
+# (Optional) The master password for the database instance. If not provided, a random password will be generated.
+#jira_rds_master_password     = "<RDS_MASTER_PASSWORD>"
```
nanux, line 88:
```suggestion
#jira_ds_master_password     = "<DB_MASTER_PASSWORD>"
```

config.tfvars
```diff
@@ -77,6 +77,16 @@ jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
 
+# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier. This will
+# restore the database from the snapshot and will not create a new database.
+#jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"
+
+# (Optional) The master username for the database instance. If you provided, the username is "postgres".
```
nanux, line 84:
```suggestion
# (Optional) The master username for the database instance. If not provided, the username is "postgres".
```

config.tfvars
```diff
@@ -77,6 +77,16 @@ jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
 
+# If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier. This will
```
nanux, line 80: I think we should include a new "Section" for the following three parameters with it's own comment.

Something like:

```
####### Dataset restore
To restore the dataset into the newly created instance, uncomment the following lines and provide all necessary parameters. Please follow XY documentation to see the requirements for dataset restoration.
```

modules/products/jira/locals.tf
```diff
@@ -40,4 +40,10 @@ locals {
       tag = var.version_tag
     }
   }) : yamlencode({})
+
+  ignore_index_check = var.db_snapshot_identifier != null ? yamlencode({
```
nanux, line 44: It would be good to add a comment why we do this and link the KB article.

variables.tf
```diff
@@ -177,6 +177,32 @@ variable "jira_db_iops" {
...
+variable "jira_db_snapshot_identifier" {
+  description = "The identifier for the DB snapshot to restore from. The snapshot should be in the same AWS Region as the DB instance."
+  default     = null
+  type        = string
+}
+
+variable "jira_rds_master_username" {
+  description = "Master username for the Jira RDS instance."
+  type        = string
+  default     = null
```
nanux, line 189: Can we make the `default = postgres` so it is explicitly defined here and easier to find? We would also save a conditional check in `locals.tf`.

variables.tf
```diff
@@ -177,6 +177,32 @@ variable "jira_db_iops" {
...
+  validation {
+    condition     = can(regex("^[a-zA-Z_]([a-zA-Z0-9_]).{5,30}$", var.jira_rds_master_username)) || var.jira_rds_master_username == null
+    error_message = "Master username must be set. It must be between 6 and 31 characters long and start with a letter/underscore and contain combination of numbers, letters, and underscore."
+  }
+}
+
+variable "jira_rds_master_password" {
+  description = "Master password for the Jira RDS instance."
```
nanux, line 197:
```suggestion
  description = "Master password for the Jira RDS instance. If not provided, the password will be generated randomly."
```

modules/AWS/rds/main.tf
```diff
@@ -40,20 +40,29 @@ module "db" {
...
 
   subnet_ids             = var.vpc.private_subnets
   vpc_security_group_ids = [module.security_group.security_group_id]
 
   maintenance_window = "Mon:00:00-Mon:03:00"
   backup_window      = "03:00-06:00"
 
+  # Snapshot settings
+  snapshot_identifier         = var.snapshot_identifier
+  allow_major_version_upgrade = true
```
nanux, line 56: Is this necessary for snapshot restore?

docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
```diff
@@ -89,3 +89,29 @@ jira_db_iops = 1000
 
 !!! info "The allowed value range of IOPS may vary based on instance class"
 You may want to adjust these values according to your needs. For more information, see [Amazon RDS DB instance storage — Amazon Relational Database Service](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html){.external}.
+
+### Database Snapshot Identifier
```
nanux, line 93: I think **Dataset restore** deserves its own section.

config.tfvars
nanux: Or if we want to use the `rds` prefix, then I think we should also use `jira_rds_snapshot_identifier`. I do prefer the `db` prefix because it is less prescriptive.

nanux:
Kicked off End to End tests with this branch.

modules/AWS/rds/main.tf
```diff
@@ -40,20 +40,29 @@ module "db" {
...
   backup_retention_period = 0
 
   skip_final_snapshot = true
   apply_immediately   = true
 }
+
+resource "random_password" "password" {
+  length           = 12
+  special          = true
+  override_special = "!@#$%^&*(){}?<>,."
```
yzha645, line 67: `@` should not be in the list. 
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html 
> The password for the database master user can include any printable ASCII character except /, ", @, or a space. Master password length constraints differ for each database engine. For more information, see the available settings when creating each DB instance.

And in variable validation checking as well.

modules/AWS/rds/variables.tf
```diff
@@ -45,3 +45,21 @@ variable "major_engine_version" {
     error_message = "Invalid major engine version. Valid ranges are from 10 to 13 (integer)."
   }
 }
+
+variable "snapshot_identifier" {
+  description = "Snapshot identifier for RDS. If specified, the DB instance is created from the snapshot."
```
yzha645, line 50:
```suggestion
  description = "Snapshot identifier for RDS. If specified, the DB instance will be created from this snapshot."
```

modules/products/jira/variables.tf
```diff
@@ -82,3 +82,22 @@ variable "pvc_claim_name" {
     error_message = "Invalid claim name."
   }
 }
+
+variable "db_snapshot_identifier" {
+  description = "Snapshot identifier for RDS. The snapshot should be in the same AWS Region as the DB instance."
```
yzha645, line 87:
```suggestion
  description = "Snapshot identifier for RDS. The snapshot should be in the same AWS region as the DB instance."
```

yzha645 (changes requested):
Do you think it's worth adding unit test for newly introduced variables as well?

variables.tf
nghazali: We can, but I this default is for postgres. This is true that postgres is the only supported DB but if we introduce any other database in future, they will have different defaults. So I moved the default to the RDS module instead. In config file is documented about default if username is not provided.

variables.tf
nghazali: I prefer to stick to our pattern to provide all additional information including the default values in comments in the config template file and documentation page. In variable definition we just describe what the variable is for. 

We have this pattern for other variables as well. At the end of the day, variable files is not a reference for users, config template and documentations are.

modules/AWS/rds/main.tf
nghazali: This is needed to avoid put additional restriction to match the major version of database with what is in snapshot. 
The process of restore the data from snapshot is first terraform provisions the database instance, and then restores snapshot. If the major version is not matched, then terraform tries to match them with upgrade/downgrade the major engine version. IF this is not possible, then an error will be raised follow by install interruption. 

To avoid it, we can set this flag true and let RDS match the versions. However, I feel we can do it only when snapshot is provided to make the impact more limited. Something like:
```
allow_major_version_upgrade = var.snapshot_identifier != null
```

docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
```diff
@@ -89,3 +89,29 @@ jira_db_iops = 1000
...
+
+`jira_db_snapshot_identifier` sets the identifier for the DB snapshot to restore from. If you do not specify a value, no AWS RDS snapshot is used.
+
+```terraform
+jira_db_snapshot_identifier = "<SNAPSHOT_IDENTIFIER>"   # e.g. "my-snapshot"
+```
+
+!!! info "The AWS RDS snapshot must be in the same Region and account as the RDS instance. If you are restoring from a shared RDS snapshot, the RDS instance must have backup retention enabled."
```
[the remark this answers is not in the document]
nghazali: I replaced it by a relevant info:
```
You also need to provide the master user credentials (`jira_db_master_username` and `jira_db_master_password`) that match the snapshot.
```

yzha645:
> Do you think it's worth adding unit test for newly introduced variables as well?

If it'll not be addressed in this PR, can you please create another sub-task to track it? 
Otherwise, all the rest changes are good to go~ 🥇

yzha645 (approved):

nanux (approved):

modules/AWS/rds/main.tf
```diff
@@ -40,20 +40,29 @@ module "db" {
   allocated_storage = var.allocated_storage
   iops              = var.iops
 
-  name                   = var.product
-  username               = local.db_master_usr
-  create_random_password = true
-  random_password_length = 12
-  port                   = 5432
+  name     = var.rds_instance_name
```
yzha645, line 43: Do we need to pass in this variable from Jira module? With some conditional logic around if snapshot_identifier is present.

nghazali: Actually there is no extra logic needed. RDS instance name is optional value for RDS. Identifier is the mandatory one.  Default values works for both Snapshot and without the snapshot restore.

yzha645: The current solution will break for non-snapshot deployment, i.e. no db will be created.

nghazali: @yzha645 , thanks for bringing this up... you are right, we need to consider this. This is fixed now.

PR #196, opened 2022-04-07
TFKUBE-414: Added Bitbucket, Jira, and Confluence modules in cleanup script

nghazali:
## Pull request description

## Checklist


nanux (approved):

nanux:
I must confess that I don't use the cleanup script explicitly. What is its main use case and how do you use it @nghazali ?

nghazali:
@nanux This script call by installer and uninstaller scripts indirectly. Usually we don't have to use in directly.

PR #197, opened 2022-04-08
TFKUBE-401: Follow up for jira snapshot restore

nghazali:
## Pull request description

## Checklist

### DIFF
--- a/modules/AWS/rds/locals.tf
+++ b/modules/AWS/rds/locals.tf
@@ -1,7 +1,8 @@
 locals {
+  product_db_name    = var.product_db_name == null ? var.product : var.product_db_name
   db_master_username = var.db_master_username == null ? "postgres" : var.db_master_username
   db_master_password = var.db_master_password == null ? random_password.password.result : var.db_master_password
-  db_jdbc_connection = "jdbc:postgresql://${module.db.db_instance_endpoint}/${module.db.db_instance_name}"
+  db_jdbc_connection = "jdbc:postgresql://${module.db.db_instance_endpoint}/${local.product_db_name}"
 
   # RDS major version is mapped to the lastest minor version for more details.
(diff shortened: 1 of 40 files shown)


modules/AWS/rds/locals.tf
```diff
@@ -1,5 +1,7 @@
 locals {
-  product_db_name    = var.product_db_name == null ? var.product : var.product_db_name
+  product_db_name   = var.product_db_name == null ? var.product : var.product_db_name
```
yzha645, line 2: What is `local.product_db_name` used for? Seems like `var.product_db_name` never get passed in via jira module...

nghazali: I cleaned it up.

modules/AWS/rds/locals.tf
```diff
@@ -1,5 +1,7 @@
 locals {
-  product_db_name    = var.product_db_name == null ? var.product : var.product_db_name
+  product_db_name   = var.product
```
yzha645, line 2: nit: is it necessary to create a new variable for it? but I guess it clearly states that it's the db name.

modules/AWS/rds/locals.tf
```diff
@@ -1,5 +1,7 @@
...
+  rds_instance_name = var.snapshot_identifier == null ? var.product : null
```
yzha645, line 3: `rds_instance_name` will make me think it's `module.rds.identifier`. I would suggest naming the variable `db_name`.

jjeongatl (approved):

modules/AWS/rds/locals.tf
nghazali: I renamed `rds_instance_id` to `rds_instance_indentifier` in bitbucket snapshot PR to make the variable name more clear. Considering that change this name will not be confused anymore.

PR #199, opened 2022-04-12
TFKUBE-409: Confluence snapshot restore

yzha645:
## Pull request description

This PR enables Confluence to restore database from RDS snapshot.

## Checklist


yzha645:
Currently there is some issue on e2e test and makes it flaky. This is not related to this PR, but makes the latest e2e test runs fail. 
However, this PR had 2 successful runs with/without domain, so make it ready for review first. 
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2151514537
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2153052694

config.tfvars
```diff
@@ -130,6 +130,14 @@ confluence_db_instance_class       = "db.t3.micro"
 confluence_db_allocated_storage    = 100
 confluence_db_iops                 = 1000
 
+# DB restore
```
jjeongatl, line 133: Can we sync description and structure that overlaps with jira?

config.tfvars
```diff
@@ -130,6 +130,14 @@ confluence_db_instance_class       = "db.t3.micro"
...
+# If you choose to restore database from an RDS snapshot, please provide the snapshot name, as well as master username and password of the snapshot.
+# Please also provide build number stored within the snapshot and Confluence license, so that Confluence can be fully setup prior to start.
+#confluence_db_snapshot_identifier = "<DB_SNAPSHOT_IDENTIFIER>"
+#confluence_db_snapshot_build_number = "<BUILD_NUMBER>"
+#confluence_db_master_username = "<MASTER_USERNAME>"
```
jjeongatl, line 138: username and password are not snapshot specific properties. We can try split them and put in separate block

docs/docs/userguide/configuration/CONFLUENCE_CONFIGURATION.md
```diff
@@ -109,4 +109,48 @@ You may want to adjust these values according to your needs. For more informatio
...
+
+`confluence_db_master_password` sets the password for the RDS master user. If you do not specify a value, a random password will be generated.
+
+```terraform
+confluence_db_master_password = "<DB_MASTER_PASSWORD>"   # default value is null
+```
+
+### Build Number
+
+`confluence_db_snapshot_build_number` configures Confluence instance with the correct build number that stores in the snapshot.
```
jjeongatl, line 151: Maybe put the link of list of build numbers here so users can look up their products' build number

jjeongatl (changes requested):

config.tfvars
nghazali, line 138: Good to match the comments here with Jira and Bitbucket to be consistent

docs/docs/userguide/configuration/CONFLUENCE_CONFIGURATION.md
nghazali: It is worth to mention this variable is mandatory if snapshot dataset is provided and otherwise will be ignored.

modules/products/confluence/variables.tf
```diff
@@ -99,4 +99,32 @@ variable "pvc_claim_name" {
...
+  type        = string
+  default     = null
+}
+
+variable "db_snapshot_build_number" {
+  description = "Confluence build number of the database snapshot."
+  type        = string
+  default     = null
+  validation {
+    condition     = var.db_snapshot_build_number == null || can(regex("^[0-9]{4}$", var.db_snapshot_build_number))
```
nghazali, line 127: Is the build number sharp 4 digits? If it can be 3 or 5 digits (builds in before or future) then better use something like `^([0-9]).{2,4}$`

modules/products/confluence/variables.tf
```diff
@@ -99,4 +99,32 @@ variable "pvc_claim_name" {
...
+    error_message = "Build number must be provided in the right format when using snapshot to restore database."
```
nghazali, line 128: If the build number is exact 4 digit, then we need to mention in validation error message. This message is not clear enough.

nghazali (approved):

modules/products/confluence/variables.tf
yzha645: It's 4 digits at least from Confluence 7.13 (minimum supported version of Helm chart). 
https://developer.atlassian.com/server/confluence/confluence-build-information/ 
And I don't see it bumping to 5 digits in the near future, so will keep it as 4-digit for now. But will make the validation message more explicit as you mentioned below.

docs/docs/userguide/configuration/CONFLUENCE_CONFIGURATION.md
yzha645: @nghazali Modified line 115 to indicate all the parameters in 'Database restore configuration' section are mandatory.

docs/docs/userguide/configuration/CONFLUENCE_CONFIGURATION.md
```diff
@@ -109,4 +109,50 @@ You may want to adjust these values according to your needs. For more informatio
 
 ```terraform
 confluence_collaborative_editing_enabled = true
+```
+
+## Dataset restore configuration
```
jjeongatl, line 114:
```suggestion
## Database restore configuration
```

jjeongatl (approved):

modules/products/confluence/variables.tf
nghazali: If the documentation says at least 4 digit so perhaps we are safe to add this validation:
`^([0-9]).{3,4}$` to cover 4 digits and 5 digits for future and copy the validation message from confluence documentation.

PR #202, opened 2022-04-14
TFKUBE-412: Add shared home restore

nanux:
## Pull request description

* Moved the NFS helm chart to the Terraform project for flexibility reasons
* Add Bitbucket snapshot ID variable (defaults to `null`)
* Expanded the NFS helm chart to include an option to define custom volume claim
* EKS module now returns the single availability zone that is used for the EKS EC2 nodes
* Bitbucket module now explicitly creates the EBS volume (from snapshot if provided), creates PV from it and then PVC that is mounted into NFS server as a shared home
* Dropped `subpath` for the Bitbucket Server helm chart shared home value
* Fixed the E2E test that is verifying a file being written to shared home
* Increased Bitbucket timeout (this should be converted to a variable)
* Increased the default NFS server resources

## Todo
- [x] fix unit tests
- [x] add variable to `config.tfvars`

## Checklist

### DIFF
--- a/dc-infrastructure.tf
+++ b/dc-infrastructure.tf
@@ -191,4 +191,6 @@ module "bitbucket" {
   elasticsearch_mem      = var.bitbucket_elasticsearch_mem
   elasticsearch_storage  = var.bitbucket_elasticsearch_storage
   elasticsearch_replicas = var.bitbucket_elasticsearch_replicas
+
+  shared_home_snapshot_id = var.bitbucket_shared_home_snapshot_id
 }
--- a/modules/AWS/eks/outputs.tf
+++ b/modules/AWS/eks/outputs.tf
@@ -38,3 +38,7 @@ output "max_cluster_size" {
 output "min_cluster_size" {
   value = module.eks.node_groups["appNodes"]["scaling_config"][0]["min_size"]
 }
+
+output "availability_zone" {
+  value = data.aws_subnet.eks_subnet.availability_zone
+}
--- a/modules/AWS/eks/outputs.tf
+++ b/modules/AWS/eks/outputs.tf
@@ -40,5 +40,7 @@ output "min_cluster_size" {
 }
 
 output "availability_zone" {
-  value = data.aws_subnet.eks_subnet.availability_zone
+  # TODO we are assuming the used availability zone is always aws-region"a"
+  # Fetching this value dynamically from ASG subnet brings additional challenges
+  value = "${var.region}a"
 }
--- a/modules/products/bitbucket/helm.tf
+++ b/modules/products/bitbucket/helm.tf
@@ -60,6 +60,11 @@ resource "helm_release" "bitbucket" {
           persistentVolumeClaim = {
             create           = true
             storageClassName = ""
+            resources = {
+              requests = {
+                storage = var.shared_home_size
+              }
+            }
           }
         }
       }
--- a/modules/products/bitbucket/helm.tf
+++ b/modules/products/bitbucket/helm.tf
@@ -6,7 +6,7 @@ resource "helm_release" "bitbucket" {
   repository = local.helm_chart_repository
   chart      = local.product_name
   version    = local.bitbucket_helm_chart_version
-  timeout    = 10 * 60 # autoscaler potentially needs to scale up the cluster
+  timeout    = 120 * 60 # TODO - convert this to variable with default of 10 minutes
 
   values = [
     yamlencode({
@@ -61,7 +61,6 @@ resource "helm_release" "bitbucket" {
             create           = true
             storageClassName = ""
           }
-          subPath = "${local.product_name}-${random_string.random.result}"
         }
       }
     }),
@@ -80,10 +79,4 @@ data "kubernetes_service" "bitbucket" {
     name      = local.product_name
     namespace = var.namespace
   }
-}
-
-resource "random_string" "random" {
-  length  = 10
-  special = false
-  number  = true
 }
--- a/modules/products/bitbucket/helm.tf
+++ b/modules/products/bitbucket/helm.tf
@@ -60,11 +60,6 @@ resource "helm_release" "bitbucket" {
           persistentVolumeClaim = {
             create           = true
             storageClassName = ""
-            resources = {
-              requests = {
-                storage = var.shared_home_size
-              }
-            }
           }
         }
       }
--- a/modules/products/bitbucket/nfs/main.tf
+++ b/modules/products/bitbucket/nfs/main.tf
@@ -1,8 +1,8 @@
 resource "aws_ebs_volume" "shared_home" {
   availability_zone = var.availability_zone
 
-  snapshot_id = var.shared_home_snapshot_id ? var.shared_home_snapshot_id : null
-  size        = var.capacity
+  snapshot_id = var.shared_home_snapshot_id != null ? var.shared_home_snapshot_id : null
+  size        = tonumber(regex("\\d+", var.capacity))
   type        = "gp2"
 
   tags = {
@@ -18,7 +18,7 @@ resource "kubernetes_persistent_volume" "shared_home" {
   spec {
     access_modes = ["ReadWriteOnce"]
     capacity = {
-      storage = "${var.capacity}G"
+      storage = var.capacity
     }
     storage_class_name = "gp2"
     persistent_volume_source {
@@ -38,7 +38,7 @@ resource "kubernetes_persistent_volume_claim" "shared_home" {
     access_modes = ["ReadWriteOnce"]
     resources {
       requests = {
-        storage = "${var.capacity}G"
+        storage = var.capacity
       }
     }
     storage_class_name = "gp2"
--- a/modules/products/bitbucket/nfs/main.tf
+++ b/modules/products/bitbucket/nfs/main.tf
@@ -13,7 +13,7 @@ resource "aws_ebs_volume" "shared_home" {
 
 resource "kubernetes_persistent_volume" "shared_home" {
   metadata {
-    name = "bitbucket-shared-home-pv"
+    name = "bitbucket-nfs-pv"
   }
   spec {
     access_modes = ["ReadWriteOnce"]
@@ -31,7 +31,7 @@ resource "kubernetes_persistent_volume" "shared_home" {
 
 resource "kubernetes_persistent_volume_claim" "shared_home" {
   metadata {
-    name      = "bitbucket-shared-home-pvc"
+    name      = "bitbucket-nfs-pvc"
     namespace = var.namespace
   }
   spec {
--- a/modules/products/bitbucket/nfs/main.tf
+++ b/modules/products/bitbucket/nfs/main.tf
@@ -1,5 +1,54 @@
+resource "aws_ebs_volume" "shared_home" {
+  availability_zone = var.availability_zone
+
+  snapshot_id = var.shared_home_snapshot_id ? var.shared_home_snapshot_id : null
+  size        = var.capacity
+  type        = "gp2"
+
+  tags = {
+    Name = "nfs-shared-home"
+  }
+}
+
+
+resource "kubernetes_persistent_volume" "shared_home" {
+  metadata {
+    name = "bitbucket-shared-home-pv"
+  }
+  spec {
+    access_modes = ["ReadWriteOnce"]
+    capacity = {
+      storage = "${var.capacity}G"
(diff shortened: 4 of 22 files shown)


modules/products/bitbucket/helm.tf
```diff
@@ -6,7 +6,7 @@ resource "helm_release" "bitbucket" {
   repository = local.helm_chart_repository
   chart      = local.product_name
   version    = local.bitbucket_helm_chart_version
-  timeout    = 10 * 60 # autoscaler potentially needs to scale up the cluster
+  timeout    = 120 * 60 # TODO - convert this to variable with default of 10 minutes
```
nghazali, line 9: Or maybe 10 minutes for default replica, and for each extra pod add 2 min?

modules/products/bitbucket/nfs/main.tf
```diff
@@ -1,13 +1,62 @@
+resource "aws_ebs_volume" "shared_home" {
```
nghazali, line 1: Are we going to use nfs for share home of all products? in this case I think better we move nfs out of bitbucket module put in in `modules/AWS`

test/e2etest/bitbucket_test.go
```diff
@@ -41,7 +41,7 @@ func assertBitbucketNfsConnectivity(t *testing.T, testConfig TestConfig) {
 	returnCode, kubectlError := k8s.RunKubectlAndGetOutputE(t, kubectlOptions,
 		"exec", "bitbucket-nfs-server-0",
 		"--", "/bin/bash",
-		"-c", "echo \"Greetings from an NFS\" >> $(find /srv/nfs/bitbucket-* | head -1)/nfs-file-share-test.txt; echo $?")
+		"-c", "echo \"Greetings from an NFS\" >> $(find /srv/nfs/ | head -1)/nfs-file-share-test.txt; echo $?")
```
nanux, line 44: We don't use the `subpath` for shared home anymore, therefore we are expecting the shared home content to be in the root of the nfs shared home volume.

modules/products/bitbucket/variables.tf
```diff
@@ -172,4 +172,14 @@ variable "db_master_password" {
...
   type        = string
   default     = null
+}
+
+variable "shared_home_snapshot_id" {
+  description = "EBS Snapshot ID with shared home content."
+  type        = string
+  default     = null
+  validation {
+    condition     = var.shared_home_snapshot_id == null || can(regex("^.{1,255}$", var.shared_home_snapshot_id))
```
nanux, line 182:
```suggestion
    condition     = var.shared_home_snapshot_id == null || can(regex("^snap-\\w{17}$", var.shared_home_snapshot_id))
```

variables.tf
```diff
@@ -483,6 +483,15 @@ variable "bitbucket_elasticsearch_replicas" {
   type        = number
   default     = 2
 }
+variable "bitbucket_shared_home_snapshot_id" {
+  description = "EBS Snapshot ID with shared home content."
+  type        = string
+  default     = null
+  validation {
+    condition     = var.bitbucket_shared_home_snapshot_id == null || can(regex("^.{1,255}$", var.bitbucket_shared_home_snapshot_id))
```
nanux, line 491:
```suggestion
    condition     = var.bitbucket_shared_home_snapshot_id == null || can(regex("^snap-\\w{17}$", var.bitbucket_shared_home_snapshot_id))
```

dc-infrastructure.tf
```diff
@@ -194,4 +194,6 @@ module "bitbucket" {
   elasticsearch_mem      = var.bitbucket_elasticsearch_mem
   elasticsearch_storage  = var.bitbucket_elasticsearch_storage
   elasticsearch_replicas = var.bitbucket_elasticsearch_replicas
+
+  shared_home_snapshot_id = var.bitbucket_shared_home_snapshot_id
```
nanux, line 198: This variable will need to be added to all products (`confluence_shared_home_snapshot_id`, `jira_shared_home_snapshot_id`)

modules/products/bitbucket/helm.tf
nanux: I would try to avoid being explicitly smart regarding timeout - there are too many variables that will impact this. Pod resources, EC2 node capacity in the EKS cluster, snapshot restore.

Exposing it as a variable would provide enough flexibility to solve the problem (the dc apps config can then define the config based on the other parameters and actual tests).

modules/products/bitbucket/helm.tf
```diff
@@ -61,7 +61,6 @@ resource "helm_release" "bitbucket" {
             create           = true
             storageClassName = ""
           }
-          subPath = "${local.product_name}-${random_string.random.result}"
```
nanux, line 64: We don't use `subpath` anymore as this is useful for EFS that is shared across products but not necessarily for the NFS scenario.

modules/products/bitbucket/helm.tf
nghazali: As far as we won't add this variable (because of time constrain in this project), we may can add timeout when we have more than one replica make some buffer and decrease the chance of failure until we refactor this part in future if is needed. 
Something like:
```
timeout    = (8 + 2 * var.replica_count) * 60  # autoscaler potentially needs to scale up the cluster
```

modules/AWS/eks/outputs.tf
```diff
@@ -38,3 +38,9 @@ output "max_cluster_size" {
 output "min_cluster_size" {
   value = module.eks.node_groups["appNodes"]["scaling_config"][0]["min_size"]
 }
+
+output "availability_zone" {
+  # TODO we are assuming the used availability zone is always aws-region"a"
```
nghazali, line 43: Do we need to refactor now, or if not how much this assumption is valid?

modules/products/bitbucket/nfs/variables.tf
```diff
@@ -18,23 +18,38 @@ variable "capacity" {
 variable "requests_cpu" {
   description = "The minimum CPU compute to request for the NFS instance"
   type        = string
-  default     = "0.25"
+  default     = "1"
```
nghazali, line 21: I think we don't have to change the default resources for NFS module. These resources could be overwritten by config file if is needed such as all other resources.

modules/products/bitbucket/variables.tf
nghazali: Is this suggestion because of the snapshot name created in AWS? Can user customise the snapshot name, for instance if create it by CLI or any other tool?

variables.tf
jjeongatl: Should Every snapshot start with `snap-`?

modules/products/bitbucket/nfs/nfs-server/templates/NOTES.txt
```diff
@@ -0,0 +1 @@
+NFS Server installed.
```
jjeongatl, line 1: What is this for?

modules/products/bitbucket/nfs/main.tf
nanux: We will do it as part of other story.

test/e2etest/bitbucket_test.go
nghazali: This can be okay based on the assumption that re-installing the nfs will move the data from last installation. I personally don't expect the data is persistent after nfs reinstall, but it worth to test?

modules/products/bitbucket/nfs/nfs-server/templates/NOTES.txt
nanux: The `NOTES.txt` file is printed after the helm chart installation happens.

modules/products/bitbucket/nfs/variables.tf
jjeongatl: Wasn't 0.25 insufficient even as a default value?

yzha645:
Just a question regarding ebs driver. You mentioned before that we need to manually install the driver. Shall we include as part of this PR? Or we can do it as part of [TFKUBE-426](https://bulldog.internal.atlassian.com/browse/TFKUBE-426). 

Did a quick search, seems like it's always a [default plugin of Kubernetes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes), but only recently made GA as [an AWS EKS addon](https://aws.amazon.com/blogs/containers/amazon-ebs-csi-driver-is-now-generally-available-in-amazon-eks-add-ons/). Latest Terraform EKS module supports `cluster_addons` we might able to utilise, but not within version 17.24.0.

jjeongatl (approved):

test/e2etest/bitbucket_test.go
nanux: The shared home and its volume are destroyed. I have tested that. Local home still hangs around, but that was the case before as well.

modules/products/bitbucket/variables.tf
nanux: No, it is always this snapshot ID, there is no other accepted format in the terraform module. I have based this regex on the snapshots across regions and old public ones. Seems that the format didn't change and it is always similar to: `snap-0c00767b17c3cca7f`

variables.tf
nanux: Yes, see https://github.com/atlassian-labs/data-center-terraform/pull/202#discussion_r852535213

modules/AWS/eks/outputs.tf
nanux: This assumption is valid. It is not a straight refactor. I had a different solution in place but it wasn't testable with `terratest`. I think it is good for now and this PR.

modules/products/bitbucket/nfs/variables.tf
nanux: I think 0.25 is just too small - even with a single product installed, we have the capacity to increase this to 0.5 or 1. We definitely need to increase the `limits` as the current values will cause NFS server to be restored (and I assume result into flakiness of the product).

nanux:
E2E tests passing:
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2187073951
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2187074348

modules/products/bitbucket/helm.tf
nanux: But that doesn't take into account resources or even if the snapshot is restored. Exposing this variable is quite simple I just didn't want to inflate the PR.

modules/products/bitbucket/nfs/main.tf
```diff
@@ -1,13 +1,62 @@
...
+  availability_zone = var.availability_zone
+
+  snapshot_id = var.shared_home_snapshot_id != null ? var.shared_home_snapshot_id : null
+  size        = tonumber(regex("\\d+", var.capacity))
+  type        = "gp2"
```
nanux, line 6: - [ ] extract `storageClass` to local variables

modules/products/bitbucket/nfs/nfs-server/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+name: nfs-server
+description: A NFS server for testing purposes. Do not use in production.
+
+type: application
+version: 0.1.0
```
nanux, line 6: - [ ] bump this version to `0.2.0` as it is different to the one that we have in repository

nanux:
E2E tests passing:
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2187073951
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2187074348

nanux:
E2E tests passing:
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2187073951
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2187074348

yzha645 (approved):

nanux:
> Did a quick search, seems like it's always a [default plugin of Kubernetes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes), but only recently made GA as [an AWS EKS addon](https://aws.amazon.com/blogs/containers/amazon-ebs-csi-driver-is-now-generally-available-in-amazon-eks-add-ons/). Latest Terraform EKS m

The implementation doesn't rely on EBS CSI driver being installed. The EBS can be used as PV even in default EKS configuration. And because we are creating the EBS volume from the snapshot outside of the cluster (in Terraform), this works even with the addon.

PR #205, opened 2022-04-19
TFKUBE-433: Extract NFS module and move to AWS modules folder

nghazali:
## Pull request description

## Checklist


modules/AWS/nfs/main.tf
```diff
@@ -6,14 +6,13 @@ resource "aws_ebs_volume" "shared_home" {
   type        = local.storage_class
 
   tags = {
-    Name = "${local.product}-nfs-shared-home"
+    Name = "nfs-shared-home"
```
yzha645, line 9: Any reason we're removing product from the tag?

yzha645 (approved):

modules/AWS/nfs/main.tf
nghazali: Right, it should be change to `${var.product}-nfs-shared-home`

PR #207, opened 2022-04-21
TFKUBE-422: Jira NFS

jjeongatl:
## Pull request description

## Checklist


modules/AWS/rds/locals.tf
```diff
@@ -1,6 +1,4 @@
 locals {
-  db_name = var.snapshot_identifier == null ? var.product : null
```
jjeongatl, line 2: Removing this condition to avoid confusion. We should recommend users to take snapshot from terraform environment

modules/AWS/rds/main.tf
```diff
@@ -40,7 +40,7 @@ module "db" {
   allocated_storage = var.allocated_storage
   iops              = var.iops
 
-  name     = local.db_name
```
jjeongatl, line 43: This means rds snapshot now should have product name as rds db_name

modules/products/jira/main.tf
```diff
@@ -28,3 +42,41 @@ module "database" {
   db_master_username      = var.db_master_username
   db_master_password      = var.db_master_password
 }
+
+resource "kubernetes_persistent_volume" "share-home-pv" {
```
jjeongatl, line 46: We could potentially move PV and PVC to NFS module and expose PVC name

jjeongatl:
E2E TEST
Domain: https://github.com/atlassian-labs/data-center-terraform/actions/runs/2200392979

nghazali:
Have you tested the solution when Bitbucket and Jira are installed at the same time? If not, please add `bitbucket` to the application list in config file and install it to see if two NFS instance can work at the same time and applications can run as expected?

modules/products/jira/helm.tf
```diff
@@ -52,10 +52,9 @@ resource "helm_release" "jira" {
         sharedHome = {
           customVolume = {
             persistentVolumeClaim = {
-              claimName = var.pvc_claim_name
+              claimName = kubernetes_persistent_volume_claim.share-home-pvc.metadata.0.name
             }
           }
-          subPath = "${local.product_name}-${random_string.random.result}"
```
nghazali, line 58: Not sure, I feel we kind of need to keep subPath at a point. Let's first test running Jira and Bitbucket together in the same environment while both using `nfs` for shared home to make sure if we have two nfs instances, they don't share the volume. I tested a couple of days ago and I faced some problem. I had not enough time to investigate more. It may was some other issue.

variables.tf
```diff
@@ -221,6 +221,46 @@ variable "jira_db_master_password" {
...
+variable "jira_shared_home_size" {
+  description = "Storage size for Jira shared home"
+  type        = string
+  default     = "10Gi"
+}
+
+variable "jira_nfs_requests_cpu" {
+  description = "The minimum CPU compute to request for the NFS instance"
+  type        = string
+  default     = "0.25"
```
nghazali, line 233: We may need to makes the default value for nfs resources consistent with Bitbucket. Unless we have a reason why those are different.

nghazali (changes requested):

modules/products/jira/main.tf
nghazali: That's a good idea, I already done this refactor partially before. Actually I created a utility module to get the nfs server ip and create a PV and PVC with requested size. I can complete it and raise a PR.

modules/AWS/rds/locals.tf
jjeongatl: Thats not bad idea, I removed this condition because if users take snapshot from terraform environment the snapshot will have product name as db_name and there is no way to workaround it. I think it is better let user configure it

variables.tf
jjeongatl: Which property has inconsistent value? I checked and they all same

config.tfvars
```diff
@@ -64,8 +64,22 @@ jira_min_heap            = "384m"
...
+# Shared home restore configuration
+# To restore shared home dataset, you can provide EBS snapshot ID that contains content of the shared home volume.
+# This volume will be mounted to the NFS server and used when the product is started.
+# Make sure the snapshot is available in the region you are deploying to and it follows all product requirements.
+#jira_shared_home_snapshot_id = "<SHARED_HOME_EBS_SNAPSHOT_IDENTIFIER>"
+
 # Storage
+# initial volume size of local/share home EBS.
 jira_local_home_size = "10Gi"
+jira_shared_home_size = "10Gi"
```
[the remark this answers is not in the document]
jjeongatl: This is same as bitbucket share home size. Unless if you think it is too much/less?

modules/products/jira/main.tf
errcode1202: I think we should get @nghazali PR merged in first and then modify this PR to use it.

config.tfvars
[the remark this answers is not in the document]
nghazali: I really don't know... this should be asked from `dc-app` team or making assumption. The main strategy they made initially (for Bamboo) was the default resources should be minimum. However, for this metric I don't think is very sensitive because this much storage is not expensive in compare with ec2/db instance type and other computational resources we have.

modules/products/jira/helm.tf
nanux: If we have a different NFS server for each product (decision we can revisit later), I don't think it is a problem as the volumes are separate.

We still need to retest the products after they are all on NFS servers.

modules/products/jira/main.tf
```diff
@@ -29,3 +44,41 @@ module "database" {
   db_master_password      = var.db_master_password
   db_name                 = var.db_name
 }
+
+# resource "kubernetes_persistent_volume" "share-home-pv" {
```
nanux, line 48: Do we need this section or not?

nanux:
> Have you tested the solution when Bitbucket and Jira are installed at the same time? If not, please add `bitbucket` to the application list in config file and install it to see if two NFS instance can work at the same time and applications can run as expected?

I think we should test this explicitly but isn't this covered by the end to end test as well?

modules/products/jira/helm.tf
nghazali: Agree... I found a bug in NFS and it may caused the problem. However let's test it again when PR#211 merged.

modules/products/jira/main.tf
nghazali: Not any more. This can be deleted.

yzha645 (approved):

errcode1202:
> > Have you tested the solution when Bitbucket and Jira are installed at the same time? If not, please add `bitbucket` to the application list in config file and install it to see if two NFS instance can work at the same time and applications can run as expected?
> 
> I think we should test this explicitly but isn't this covered by the end to end test as well?

The `e2e` tests install all products so it should be covered.

modules/products/jira/variables.tf
```diff
@@ -67,27 +67,59 @@ variable "jira_configuration" {
   }
 }
 
+variable "nfs_requests_cpu" {
+  description = "The minimum CPU compute to request for the NFS instance"
+  type        = string
+  default     = "1"
```
nanux, line 73: Do we need to include defaults on this level as well if they are the same as the defaults in the `AWS/nfs` module?

config.tfvars
```diff
@@ -64,8 +64,22 @@ jira_min_heap            = "384m"
...
 jira_reserved_code_cache = "512m"
 
+# Jira NFS instance resource configuration
+#jira_nfs_requests_cpu    = "<REQUESTS_CPU>"
+#jira_nfs_requests_memory = "<REQUESTS_MEMORY>"
+#jira_nfs_limits_cpu      = "<LIMITS_CPU>"
+#jira_nfs_limits_memory   = "<LIMITS_MEMORY>"
+
+# Shared home restore configuration
+# To restore shared home dataset, you can provide EBS snapshot ID that contains content of the shared home volume.
```
errcode1202, line 74:
```suggestion
# To restore a shared home dataset, you can provide an EBS snapshot ID that contains the content of the shared home volume.
```

variables.tf
```diff
@@ -227,6 +227,46 @@ variable "jira_db_master_password" {
   }
 }
 
+variable "jira_shared_home_size" {
+  description = "Storage size for Jira shared home"
+  type        = string
+  default     = "10Gi"
```
nanux, line 233: Again, seems we have defaults on all 3 levels. Is that something we want? Would it be enough to have just 1 level with defaults (e.g. the top one)?

docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
```diff
@@ -142,3 +155,11 @@ jira_db_master_username = "<DB_MASTER_USERNAME>"   # e.g. "postgres"
 ```terraform
 jira_db_master_password = "<DB_MASTER_PASSWORD>"   # default value is null
 ```
+### Shared Home Restore
+
+`jira_shared_home_snapshot_id` sets id of shared home EBS snapshot. This will spin up EBS volume and it will be mounted to the NFS server and used when the product is started.
```
errcode1202, line 160:
```suggestion
`jira_shared_home_snapshot_id` sets the id of the shared home EBS snapshot to use. This will spin up an EBS volume that will be mounted to the NFS server and used when the product is started.
```

nanux (approved):

errcode1202 (approved):

nghazali (approved):

PR #209, opened 2022-04-21
TFKUBE-426:  Refactor NFS module - Add PV and PVC to nfs module

nghazali:
## Pull request description

Refactored the NFS module
- Added `shared_home_size` variable to the nfs module
- Added PV and PVC into nfs module (they will be created if `shared_home_size` is provided
- Moved `nfs-server` helm chart into a separated folder in root named `helm-charts`
  The helm chart code is not part of terraform so it's better to keep separated in my opinion.

## Checklist


modules/AWS/nfs/outputs.tf
```diff
@@ -1,3 +1,7 @@
 output "helm_release_nfs_service_ip" {
   value = data.kubernetes_service.nfs.spec[0].cluster_ip
+}
+
+output "nfs_pvc_claim_name" {
```
errcode1202, line 5: maybe simply `nfs_claim_name`

```suggestion
output "nfs_claim_name" {
```

errcode1202 (approved):

yzha645:
Sorry, I might need more context on the design here. Does `shared_home_size` serve the same purpose as `capacity`? These two seems duplicated to me. And we're using `capacity` in all product modules already.

nghazali:
I am merging this PR as result of our discussion in zoom call with Yifei, Jun and Dylan.

PR #210, opened 2022-04-22
Add DB_NAME to config file

jjeongatl:
## Pull request description

Previously we implemented the logic to dynamically decide db_name for RDS depending on whether db_snapshot_identifier was used or not. This is because of inconsistency between snapshot provided by dc-apt team and snapshot taken by the data-center-terraform. 
This change will allow users to use input whichever db_name matching with their RDS snapshot
## Checklist

### DIFF
--- a/config.tfvars
+++ b/config.tfvars
@@ -76,6 +76,8 @@ jira_db_major_engine_version = "12"
 jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
+# IF you want to restore the database from a snapshot, please provide the default db name for the snapshot. If the snapshot does not have default db name comment out this variable.
+jira_db_name = "jira"
 
 # Database restore configuration
 # If you want to restore the database from a snapshot, uncomment the following line and provide the snapshot identifier.
@@ -132,6 +134,8 @@ confluence_db_major_engine_version = "11"
 confluence_db_instance_class       = "db.t3.micro"
 confluence_db_allocated_storage    = 100
 confluence_db_iops                 = 1000
+# IF you want to restore the database from a snapshot, please provide the default db name for the snapshot. If the snapshot does not have default db name comment out this variable.
+confluence_db_name = "confluence"
 
 # Database restore configuration
 # If you want to restore the database from a snapshot, uncomment the following lines and provide the snapshot identifier.
@@ -206,6 +210,8 @@ bitbucket_db_major_engine_version = "13"
 bitbucket_db_instance_class       = "db.t3.micro"
 bitbucket_db_allocated_storage    = 100
 bitbucket_db_iops                 = 1000
+# IF you want to restore the database from a snapshot, please provide the default db name for the snapshot. If the snapshot does not have default db name comment out this variable.
+bitbucket_db_name = "bitbucket"
 
 # Bitbucket NFS instance resource configuration
 #bitbucket_nfs_requests_cpu    = "<REQUESTS_CPU>"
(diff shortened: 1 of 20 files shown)


config.tfvars
```diff
@@ -76,6 +76,8 @@ jira_db_major_engine_version = "12"
 jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
+# IF you want to restore the database from a snapshot, please provide the default db name for the snapshot. If the snapshot does not have default db name comment out this variable.
```
nghazali, line 79:
```suggestion
# If you restore the database, you need to provide the db name from the snapshot. 
```

config.tfvars
```diff
@@ -132,6 +134,8 @@ confluence_db_major_engine_version = "11"
 confluence_db_instance_class       = "db.t3.micro"
 confluence_db_allocated_storage    = 100
 confluence_db_iops                 = 1000
+# IF you want to restore the database from a snapshot, please provide the default db name for the snapshot. If the snapshot does not have default db name comment out this variable.
```
nghazali, line 137:
```suggestion
# If you restore the database, you need to provide the db name from the snapshot. 
```

config.tfvars
```diff
@@ -206,6 +210,8 @@ bitbucket_db_major_engine_version = "13"
 bitbucket_db_instance_class       = "db.t3.micro"
 bitbucket_db_allocated_storage    = 100
 bitbucket_db_iops                 = 1000
+# IF you want to restore the database from a snapshot, please provide the default db name for the snapshot. If the snapshot does not have default db name comment out this variable.
```
nghazali, line 213:
```suggestion
# If you restore the database, you need to provide the db name from the snapshot. 
```

modules/products/bamboo/main.tf
```diff
@@ -24,4 +24,6 @@ module "database" {
   instance_class          = var.db_configuration["db_instance_class"]
   iops                    = var.db_configuration["db_iops"]
   vpc                     = var.vpc
+  db_name                 = var.db_configuration["db_name"]
+
```
nghazali, line 28:
```suggestion
  db_name                 = var.db_configuration["db_name"]
```

nghazali: removed an extra space line.

nghazali (approved):
Good improvement. Just left some suggestions.

variables.tf
```diff
@@ -187,6 +187,12 @@ variable "jira_db_iops" {
   type        = number
 }
 
+variable "jira_db_name" {
+  description = "The default DB name of the DB instance."
+  default     = null
+  type        = string
+}
```
nghazali, line 194: we man can add validation for db name?

variables.tf
```diff
@@ -693,6 +711,12 @@ variable "bamboo_db_iops" {
   type        = number
 }
 
+variable "bamboo_db_name" {
+  description = "The default DB name of the DB instance."
+  default     = null
+  type        = string
+}
+
```
nghazali, line 719: Do we need this for Bamboo? I think we use different approach rather than snapshot for Bamboo.

nghazali: But on the other hand, is good to be consistent.

variables.tf
jjeongatl: added in the module level

variables.tf
jjeongatl: This is optional but I added to keep it consistent

nghazali:
@jjeongatl , before merge please run e2e test and complete the PR checklist

PR #213, opened 2022-04-22
TFKUBE-403: Add support for Jira Service Management

nanux:
## Pull request description

Adds an option to change the image used for Jira deployment. This enables us to use for example https://hub.docker.com/r/atlassian/jira-servicemanagement/tags

Change of the image is currently only for Jira to enable the use case of Jira Servicemanagement. It is possible to add it for other products but I have decided to not complicate them with an unclear use case.

* added `jira_image_repository` variable to make the image configurable
* added documentation to instruct users to choose a different edition (e.g. service desk)
* fixed broken links in existing documentation

## Checklist


config.tfvars
```diff
@@ -55,6 +55,7 @@ jira_replica_count = 1
 
 # By default, Jira Software will use the version defined in the Helm chart. If you wish to override the version, uncomment
 # the following line and set the jira_version_tag to any of the versions available on https://hub.docker.com/r/atlassian/jira-software/tags
+#jira_image_repository = "atlassian/jira-software"
```
nghazali, line 58: Maybe this location is not the best to land this line. The comment on lines 57 & 58 refering to jira version tag as the `following line` which now is different variable. 

Can we move it to another line and do you think if any comment is needed for `jira_image_repository` in config file?

nghazali (changes requested):

nanux:
E2E test
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2224796611
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2224797104

jjeongatl:
Does JSM share same db schema as Jira Software and Core? Interested to know if we still can use same sql query to update base_url and license on db restore

docs/docs/userguide/configuration/JIRA_CONFIGURATION.md
```diff
@@ -20,6 +20,20 @@ For more information, see [Jira Version Tags](https://hub.docker.com/r/atlassian
 jira_version_tag = "<JIRA_VERSION_TAG>"
 ```
 
+### Jira image repository
+
+To change the Jira edition you can set a different image repository. By default, Jira Software edition is installed. 
+You need to make sure the appropriate version defined in `jira_version_tag` is available for the selected Jira edition.
+
```
nghazali, line 27: I think it would be nice if we mention the possible values for this variable here.

modules/products/jira/variables.tf
```diff
@@ -114,6 +113,12 @@ variable "shared_home_snapshot_id" {
   }
 }
 
+variable "image_repository" {
+  description = "Jira image repository"
+  type        = string
+  default     = "atlassian/jira-software"
```
nghazali, line 119: We should have validation for this as the valid values are restricted to a set.

nghazali (changes requested):

config.tfvars
```diff
@@ -57,6 +57,13 @@ jira_replica_count = 1
 # the following line and set the jira_version_tag to any of the versions available on https://hub.docker.com/r/atlassian/jira-software/tags
 #jira_version_tag = "<JIRA_VERSION_TAG>"
 
+# To select a different image repository for the Jira application, you can change following variable:
+# Official suitable values are:
+# * "atlassian/jira-software"
+# * "atlassian/jira-servicemanagement"
+# * "atlassian/jira-core"
```
yzha645, line 64: nit: It's not listed in JIRA_CONFIGURATION.md, maybe we can drop `jira-core` here, given `jira-core` does not normally being installed independently anyway.

yzha645 (approved):

nanux:
> Does JSM share same db schema as Jira Software and Core? Interested to know if we still can use same sql query to update base_url and license on db restore

Yes, I believe the core schema will be the same (you can install JSM into Jira) and we can keep the query.

modules/products/jira/variables.tf
nanux: I don't want to include validation for the set as this enables testing different images as well. I was even thinking about including this variable for other products so people can test with customized images but I don't want to expand the scope.

We have documentation for official values so users should be able to know what can be plugged here.

PR #214, opened 2022-04-25
TFKUBE-441: Compromise tagging failure in installer during e2e tests.

nghazali:
## Pull request description

Compromise tagging failure in installer during e2e tests. We observed a case of failure caused by tagging module in e2e tests. As this is not reported not in our local tests and nor from other users and testers including out blitz, and considering tagging is not essential for e2e tests because the instances are short-living, so we can compromise it in specifically e2e test. 

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -211,7 +211,7 @@ add_tags_to_asg_resources() {
   terraform -chdir="${TAG_MODULE_PATH}" init -no-color > "${LOG_TAGGING}"
   terraform -chdir="${TAG_MODULE_PATH}" apply -auto-approve -no-color "${OVERRIDE_CONFIG_FILE}" >> "${LOG_TAGGING}"
   if [ $? == 0 ]; then
-    log "Resource tags are applied to ASG and all EC2 instances."
+    log "Resource tags were applied to ASG and all EC2 instances."
   else
     log "Resource tags are not applied to ASG and all EC2 instances." "ERROR"
   fi
--- a/install.sh
+++ b/install.sh
@@ -201,12 +201,21 @@ create_update_infrastructure() {
 
 # Apply the tags into ASG and EC2 instances created by ASG
 add_tags_to_asg_resources() {
+  if [ -n "${FORCE_FLAG}" ]; then
+    log "The script is executed by end2end test. Tagging of ASG and EC2 resources failure will be compromised."
+    set +e
+  fi
   log "Tagging Auto Scaling Group and EC2 instances. It may take a few minutes. Please wait..."
   TAG_MODULE_PATH="${ROOT_PATH}/modules/AWS/asg_ec2_tagging"
 
   terraform -chdir="${TAG_MODULE_PATH}" init -no-color > "${LOG_TAGGING}"
   terraform -chdir="${TAG_MODULE_PATH}" apply -auto-approve -no-color "${OVERRIDE_CONFIG_FILE}" >> "${LOG_TAGGING}"
-  log "Resource tags are applied to ASG and all EC2 instances."
+  if [ $_ == 0 ]; then
+    log "Resource tags are applied to ASG and all EC2 instances."
+  else
+    log "Resource tags are not applied to ASG and all EC2 instances." "ERROR"
+  fi
+  set -e
 }
 
 set_current_context_k8s() {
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -201,12 +201,21 @@ create_update_infrastructure() {
...
+    log "The script is executed by end2end test. Tagging of ASG and EC2 resources failure will be compromised."
+    set +e
+  fi
   log "Tagging Auto Scaling Group and EC2 instances. It may take a few minutes. Please wait..."
   TAG_MODULE_PATH="${ROOT_PATH}/modules/AWS/asg_ec2_tagging"
 
   terraform -chdir="${TAG_MODULE_PATH}" init -no-color > "${LOG_TAGGING}"
   terraform -chdir="${TAG_MODULE_PATH}" apply -auto-approve -no-color "${OVERRIDE_CONFIG_FILE}" >> "${LOG_TAGGING}"
-  log "Resource tags are applied to ASG and all EC2 instances."
+  if [ $_ == 0 ]; then
```
nanux, line 213:
```suggestion
  if [ $? == 0 ]; then
```
I haven't really used `$_` but from reading https://unix.stackexchange.com/questions/280453/understand-the-meaning-of that is not what we want to do here. I think we need to check the exit code.

install.sh
```diff
@@ -201,12 +201,21 @@ create_update_infrastructure() {
...
+    log "Resource tags are applied to ASG and all EC2 instances."
```
nanux, line 214:
```suggestion
    log "Resource tags were applied to ASG and all EC2 instances."
```

install.sh
```diff
@@ -201,12 +201,21 @@ create_update_infrastructure() {
...
+  else
+    log "Resource tags are not applied to ASG and all EC2 instances." "ERROR"
```
nanux, line 216:
```suggestion
    log "Resource tags were not applied to ASG and all EC2 instances." "ERROR"
```

install.sh
```diff
@@ -201,12 +201,21 @@ create_update_infrastructure() {
 
 # Apply the tags into ASG and EC2 instances created by ASG
 add_tags_to_asg_resources() {
+  if [ -n "${FORCE_FLAG}" ]; then
+    log "The script is executed by end2end test. Tagging of ASG and EC2 resources failure will be compromised."
```
nanux, line 205:
```suggestion
    log "The script is executed by end2end test. Failure in tagging of ASG and EC2 resources will not cause the script to error out."
```

nanux (changes requested):

nanux:
Why do we have failures in the EC2 tagging?

nanux:
> One use case of failure in tagging was version mismatch which is not applying to e2e test. Normally we don't have problem in tagging - but as part of our investigation we observed at least one case of e2e test failed that the reason was tagging module. As we didn't find similar error in out tests, so I think is not worth to spend more time for this issue and this code will prevent e2e test failure in future.

I see. Is there a possibility that this might mask a problem with our tagging code?

nghazali:
> > One use case of failure in tagging was version mismatch which is not applying to e2e test. Normally we don't have problem in tagging - but as part of our investigation we observed at least one case of e2e test failed that the reason was tagging module. As we didn't find similar error in out tests, so I think is not worth to spend more time for this issue and this code will prevent e2e test failure in future.
> 
> I see. Is there a possibility that this might mask a problem with our tagging code?

No, this will apply only in e2e test. (Only when install is executed with FORCE_FLAG. This is off by default and is an internal flag and is not documented for vendor or any other purposed use.)

nanux:
> > > One use case of failure in tagging was version mismatch which is not applying to e2e test. Normally we don't have problem in tagging - but as part of our investigation we observed at least one case of e2e test failed that the reason was tagging module. As we didn't find similar error in out tests, so I think is not worth to spend more time for this issue and this code will prevent e2e test failure in future.
> > 
> > 
> > I see. Is there a possibility that this might mask a problem with our tagging code?
> 
> No, this will apply only in e2e test. (Only when install is executed with FORCE_FLAG. This is off by default and is an internal flag and is not documented for vendor or any other purposed use.)

I understand it is only for tagging. But if the tagging will become broken, the end to end test won't really show that failure.

nanux (approved):

nghazali:
I see your point @nanux and is a valid one. Let's discuss if it worth we merge this PR or we get more benefit of drop it.

nghazali:
Regarding the team decision I close this without merge.

PR #215, opened 2022-04-25
TFKUBE-440: Use NFS instead of EFS for shared home in Bamboo

nghazali:
## Pull request description

All other products using NFS but Bamboo. This PR is adapting Bamboo with NFS and in this case we can get rid of NFS from common modules. 

Use NFS instead of EFS for shared home in Bamboo

## Checklist

### DIFF
--- a/modules/products/bamboo/helm.tf
+++ b/modules/products/bamboo/helm.tf
@@ -46,7 +46,7 @@ resource "helm_release" "bamboo" {
         sharedHome = {
           customVolume = {
             persistentVolumeClaim = {
-              claimName = var.pvc_claim_name
+              claimName = module.nfs.nfs_claim_name
             }
           }
           subPath = local.sub_path
@@ -102,10 +102,4 @@ resource "helm_release" "bamboo_agent" {
     }),
     local.agent_version_tag,
   ]
-}
-
-resource "random_string" "random" {
-  length  = 10
-  special = false
-  number  = true
 }
--- a/modules/products/bamboo/helm.tf
+++ b/modules/products/bamboo/helm.tf
@@ -7,7 +7,7 @@ resource "helm_release" "bamboo" {
   repository = local.helm_chart_repository
   chart      = local.bamboo_helm_chart_name
   version    = local.bamboo_helm_chart_version
-  timeout    = 40 * 60 # dataset import can take a long time
+  timeout    = 10 * 60
 
   values = [
     yamlencode({
@@ -49,7 +49,6 @@ resource "helm_release" "bamboo" {
               claimName = module.nfs.nfs_claim_name
             }
           }
-          subPath = local.sub_path
         }
       }
     }),
--- a/modules/products/bamboo/locals.tf
+++ b/modules/products/bamboo/locals.tf
@@ -116,5 +116,4 @@ locals {
   }) : yamlencode({})
 
   dataset_filename = "bamboo_dataset_to_import.zip"
-  sub_path         = local.product_name
 }
--- a/modules/products/bamboo/locals.tf
+++ b/modules/products/bamboo/locals.tf
@@ -116,5 +116,5 @@ locals {
   }) : yamlencode({})
 
   dataset_filename = "bamboo_dataset_to_import.zip"
-  sub_path         = "${local.product_name}-${random_string.random.result}"
+  sub_path         = local.product_name
 }
--- a/modules/products/bamboo/main.tf
+++ b/modules/products/bamboo/main.tf
@@ -26,3 +26,18 @@ module "database" {
   vpc                     = var.vpc
   db_name                 = var.db_configuration["db_name"]
 }
+
+module "nfs" {
+  source = "../../AWS/nfs"
+
+  namespace               = var.namespace
+  product                 = local.product_name
+  requests_cpu            = var.nfs_requests_cpu
+  requests_memory         = var.nfs_requests_memory
+  limits_cpu              = var.nfs_limits_cpu
+  limits_memory           = var.nfs_limits_memory
+  capacity                = var.shared_home_size
+  availability_zone       = var.eks.availability_zone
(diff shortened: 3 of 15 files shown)


modules/products/bamboo/main.tf
```diff
@@ -26,3 +26,18 @@ module "database" {
...
+
+  namespace               = var.namespace
+  product                 = local.product_name
+  requests_cpu            = var.nfs_requests_cpu
+  requests_memory         = var.nfs_requests_memory
+  limits_cpu              = var.nfs_limits_cpu
+  limits_memory           = var.nfs_limits_memory
+  capacity                = var.shared_home_size
+  availability_zone       = var.eks.availability_zone
+  shared_home_snapshot_id = null
```
nanux, line 41: Isn't this defaulted to `null`? Can we just skip it?

test/e2etest/test-config.tfvars.tmpl
```diff
@@ -86,12 +87,14 @@ jira_reserved_code_cache = "512m"
 jira_db_instance_class       = "db.t3.micro"
 jira_db_allocated_storage    = 100
 jira_db_iops                 = 1000
+jira_db_name                 = "jira"
```
nanux, line 90: Is this fixing some e2e failure?

test/e2etest/test-config.tfvars.tmpl
```diff
@@ -26,7 +26,7 @@ resource_tags = {
 }
 
 # Instance types that is preferred for node group.
-instance_types = ["m5.xlarge"]
+instance_types = ["m5.2xlarge"]
```
nanux, line 29: This does make sense. We need to scale less but we still scale up to test the autoscaler.

modules/products/bamboo/main.tf
nghazali: We can, but having this null is more explicit and it may be more clear

test/e2etest/test-config.tfvars.tmpl
nghazali: Yes, it looks this is part of solution. I have the e2e tests green now for this branch.

modules/products/bamboo/locals.tf
```diff
@@ -116,5 +116,5 @@ locals {
   }) : yamlencode({})
 
   dataset_filename = "bamboo_dataset_to_import.zip"
-  sub_path         = "${local.product_name}-${random_string.random.result}"
+  sub_path         = local.product_name
```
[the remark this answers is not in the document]
nghazali: done

modules/products/bamboo/main.tf
nghazali: dropped this

modules/products/bamboo/helm.tf
```diff
@@ -7,7 +7,7 @@ resource "helm_release" "bamboo" {
   repository = local.helm_chart_repository
   chart      = local.bamboo_helm_chart_name
   version    = local.bamboo_helm_chart_version
-  timeout    = 40 * 60 # dataset import can take a long time
+  timeout    = 10 * 60
```
jjeongatl, line 10: Will 10mins be enough to import dataset?

variables.tf
```diff
@@ -681,6 +681,36 @@ variable "bamboo_local_home_size" {
   default     = "10Gi"
 }
 
+variable "bamboo_shared_home_size" {
+  description = "Storage size for Jira shared home"
```
yzha645, line 685:
```suggestion
  description = "Storage size for Bamboo shared home"
```

modules/products/bamboo/helm.tf
nanux, line 10: Ok, but we need to action the task to provide variables for timeout soon.

modules/products/bamboo/variables.tf
```diff
@@ -115,11 +145,15 @@ variable "local_agent_chart_path" {
   default = ""
 }
 
+# TODO: This variable was added to accept pcv claim, however, based on the current
```
nanux, line 148:
```suggestion
# TODO: This variable was added to accept PVC claim, however, based on the current
```

modules/products/bamboo/variables.tf
```diff
@@ -115,11 +145,15 @@ variable "local_agent_chart_path" {
...
+#       implementation, it is not used. This could be used in the future to override
+#       the default pcv claim (use another file system instead of creating nfs by the product module.
```
nanux, line 150:
```suggestion
#       the default PVC claim (use another file system instead of creating nfs by the product module).
```

modules/products/bamboo/variables.tf
yzha645, line 148: nit: Just for consistency reason, all other products have removed this variable. 
And typo `pcv`.

nanux (approved):

test/e2etest/test-config.tfvars.tmpl
jjeongatl: This is optional variable. In which part of the e2e test checks db_name?

modules/products/bamboo/variables.tf
nghazali: I removed the variable and comments

modules/products/bamboo/variables.tf
nghazali: removed the comment and the variable

modules/products/bamboo/helm.tf
nghazali: is more than enough... for Bamboo it takes 3 minutes max

jjeongatl (approved):

test/e2etest/test-config.tfvars.tmpl
nghazali: The default value is `null` and it means rds will be created with no database. I think we should change the default values of db_name for all products from null to the product name.

PR #217, opened 2022-04-26
TFKUBE-446: Unify NFS usage across all products

nanux:
## Pull request description

* Switch Bitbucket to use `persistentVolumeClaim` instead of accessing NFS directly
* Remove `capacity` that was causing confusion
* Remove the conditional for the shared home PV and PVC creation
* Rename variables for consistency and clarity
* Fix unit tests

## Checklist


nanux:
E2E tests
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2225288204
https://github.com/atlassian-labs/data-center-terraform/actions/runs/2225288965

modules/products/bitbucket/helm.tf
```diff
@@ -50,17 +50,11 @@ resource "helm_release" "bitbucket" {
...
-              path   = "/srv/nfs"
+          customVolume = {
+            persistentVolumeClaim = {
+              claimName = module.nfs.nfs_claim_name
             }
           }
-          persistentVolumeClaim = {
-            create           = true
-            storageClassName = ""
-          }
```
errcode1202, line 63: To be honest, this is now so much cleaner too 👍🏽

errcode1202 (approved):

modules/AWS/nfs/outputs.tf
```diff
@@ -3,5 +3,5 @@ output "helm_release_nfs_service_ip" {
 }
 
 output "nfs_claim_name" {
-  value = var.shared_home_size == null ? null : kubernetes_persistent_volume_claim.shared-home-pvc[0].metadata[0].name
+  value = var.shared_home_size == null ? null : kubernetes_persistent_volume_claim.product_shared_home_pvc.metadata[0].name
```
jjeongatl, line 6: `var.shared_home_size == null ?`
Do we still need this condition?

modules/AWS/nfs/variables.tf
```diff
@@ -58,7 +52,7 @@ variable "shared_home_snapshot_id" {
     error_message = "Provide correct EBS snapshot ID."
   }
 }
-# The size of the EBS volume to allocate for shared home. If is null no PV and PVC will be created.
+
 variable "shared_home_size" {
```
nghazali, line 56: Please remove default value as this is not valid anymore. Also add validation for shared_home please.

nghazali (changes requested):

modules/AWS/nfs/outputs.tf
nanux: We don't, good catch!

modules/AWS/nfs/variables.tf
```diff
@@ -58,9 +52,12 @@ variable "shared_home_snapshot_id" {
...
-  description = "The storage capacity to allocate to local home"
+  description = "The storage capacity to allocate to shared home"
   type        = string
-  default     = null
+  validation {
+    condition     = can(regex("^[0-9]+([gG]|Gi)$", var.shared_home_size)) || var.shared_home_size == null
```
nanux, line 60: We support only G units due to this line - https://github.com/atlassian-labs/data-center-terraform/pull/217/files#diff-26d8f7b099b5c1c21aeafca0e13eb3676ace8a8bd69e0d240a7446b8cbe8c2efR5

jjeongatl (approved):

modules/AWS/nfs/variables.tf
nghazali: Null value for `shared_home` should not be valid

nghazali (changes requested):

nghazali (approved):

modules/common/variables.tf
```diff
@@ -47,6 +47,15 @@ variable "namespace" {
   }
 }
 
+variable "shared_home_size" {
+  description = "Shared home persistent volume size."
+  type        = string
+  validation {
+    condition     = can(regex("^[0-9]+([gG]|Gi)$", var.shared_home_size)) || var.shared_home_size == null
```
nghazali, line 54: Do still we need this `null` in validation.

PR #221, opened 2022-04-29
TFKUBE-448: Add limits for the elasticsearch helm chart

nanux:
## Pull request description

This change fixes bug when the user is installing Bitbucket and also redefine the Elasticsearch requests with higher values (beyond the default `limits` values in the [Elasticsearch helm chart](https://github.com/elastic/helm-charts/blob/main/elasticsearch/values.yaml#L94)). It is not possible to have `requests` being higher than `limits` and therefore the Elasticsearch statefulSet fails to deploy.

* Added `elasticsearch_limits_memory`, `elasticsearch_limits_cpu` variables with defaults (double the `requests`).
* Renamed the elasticsearch requests to align them with the `nfs` module
* Added the variables into user documentation
* Fixed the unit tests

## Checklist

### DIFF
--- a/modules/products/bitbucket/variables.tf
+++ b/modules/products/bitbucket/variables.tf
@@ -135,13 +135,23 @@ variable "elasticsearch_endpoint" {
   default     = null
 }
 
-variable "elasticsearch_cpu" {
-  description = "Number of CPUs for elasticsearch instance."
+variable "elasticsearch_requests_cpu" {
+  description = "Number of CPUs requested for elasticsearch instance."
   type        = string
 }
 
-variable "elasticsearch_mem" {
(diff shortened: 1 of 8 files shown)


modules/products/bitbucket/variables.tf
```diff
@@ -135,13 +135,23 @@ variable "elasticsearch_endpoint" {
   default     = null
 }
 
-variable "elasticsearch_cpu" {
-  description = "Number of CPUs for elasticsearch instance."
+variable "elasticsearch_requests_cpu" {
```
nanux, line 138: Rename to make it consistent with the `nfs` module

nghazali (approved):

PR #223, opened 2022-05-03
TFKUBE-453: Add additionalJvmArgs in Jira so It can pick up any existing index snapshot in share-home

jjeongatl:
## Pull request description

Valid index snapshot for Jira node should be by default created within 24 hours prior to Jira Node creation. This is an issue as dc-apt team won't distribute new shared-home snapshot everyday.

com.atlassian.jira.startup.max.age.of.usable.index.snapshot.in.hours allows user to customise the period.

It is set to 10 years for now but when jira-server team introduce infinite flag(-1) we can update it later

## Checklist

### DIFF
--- a/modules/products/jira/locals.tf
+++ b/modules/products/jira/locals.tf
@@ -44,11 +44,12 @@ locals {
     }
   }) : yamlencode({})
 
+  # After restoring the snapshot of the Jira database, a re-index is required. To avoid interruption in the Jira
+  # service we should exclude indexing status from the health check process.
+  # For more info see: https://jira.atlassian.com/browse/JRASERVER-66970
+  ignore_index_check = var.db_snapshot_id != null ? ["-Dcom.atlassian.jira.status.index.check=false"] :[]
+
   # By default, Jira accepts an index snapshot taken within 24hours. In order to use snapshot older than 24hours we need to update following property value.
   # It is set to 10 years.
-  reuse_old_index_snapshot = var.shared_home_snapshot_id != null ? yamlencode({
-    jira = {
-      additionalJvmArgs = ["-Dcom.atlassian.jira.startup.max.age.of.usable.index.snapshot.in.hours=${local.ageOfUsableIndexSnapshot}"]
-    }
-  }) : yamlencode({})
+  reuse_old_index_snapshot = var.shared_home_snapshot_id != null ? ["-Dcom.atlassian.jira.startup.max.age.of.usable.index.snapshot.in.hours=${local.ageOfUsableIndexSnapshot}"]:[]
 }
--- a/modules/products/jira/locals.tf
+++ b/modules/products/jira/locals.tf
@@ -16,6 +16,7 @@ locals {
 
   domain_supplied     = var.ingress.outputs.domain != null ? true : false
   product_domain_name = local.domain_supplied ? "${local.product_name}.${var.ingress.outputs.domain}" : null
+  ageOfUsableIndexSnapshot = 24 * 365 * 10
 
   # ingress settings for Jira service
   ingress_settings = yamlencode({
@@ -47,7 +48,7 @@ locals {
   # It is set to 10 years.
   reuse_old_index_snapshot = var.shared_home_snapshot_id != null ? yamlencode({
     jira = {
-      additionalJvmArgs = ["-Dcom.atlassian.jira.startup.max.age.of.usable.index.snapshot.in.hours=87600"]
+      additionalJvmArgs = ["-Dcom.atlassian.jira.startup.max.age.of.usable.index.snapshot.in.hours=${local.ageOfUsableIndexSnapshot}"]
     }
   }) : yamlencode({})
 }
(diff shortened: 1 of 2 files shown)


jjeongatl:
E2E No Domain: https://github.com/atlassian-labs/data-center-terraform/actions/runs/2261808310

modules/products/jira/locals.tf
```diff
@@ -43,12 +43,11 @@ locals {
     }
   }) : yamlencode({})
 
-  # After restoring the snapshot of the Jira database, a re-index is required. To avoid interruption in the Jira
```
jjeongatl, line 46: This won't be needed. as share-home will always provide index snapshot

modules/products/jira/locals.tf
```diff
@@ -43,12 +43,11 @@ locals {
...
-  # service we should exclude indexing status from the health check process.
-  # For more info see: https://jira.atlassian.com/browse/JRASERVER-66970
-  ignore_index_check = var.db_snapshot_id != null ? yamlencode({
+  # By default, Jira accepts an index snapshot taken within 24hours. In order to use snapshot older than 24hours we need to update following property value.
+  # It is set to 10 years.
+  reuse_old_index_snapshot = var.shared_home_snapshot_id != null ? yamlencode({
     jira = {
-      additionalJvmArgs = ["-Dcom.atlassian.jira.status.index.check=false"]
```
jjeongatl, line 51: This won't be needed. as share-home will always provide index snapshot

modules/products/jira/locals.tf
```diff
@@ -43,12 +43,11 @@ locals {
...
+      additionalJvmArgs = ["-Dcom.atlassian.jira.startup.max.age.of.usable.index.snapshot.in.hours=87600"]
```
errcode1202, line 50: I dont even think Jira DC is ten years old 😂  Wouldnt a 1 year grace period be sufficient?

errcode1202 (changes requested):

modules/products/jira/locals.tf
errcode1202: It might be nice if this value was configurable. This means if there are snapshot expiry issues it can be resolved easily enough by extending the grace.

errcode1202 (changes requested):

modules/products/jira/locals.tf
jjeongatl: I agree that 10 years is very generous figure. We can ask how often dc-apt team will release the share-home snapshot and calculate number from that. Key is to put enough hours to make sure jira picks up the snapshot

errcode1202 (approved):

modules/products/jira/locals.tf
jjeongatl: It may be unnecessary to expose this value in config. If we do, vendors now have one more information to figure out. Isn't it easier if we just put enough buffer so the snapshot is always picked up?

modules/products/jira/locals.tf
nghazali: What if user provide the RDS snapshot but not EBS snapshot? I think is safe to leave it as it is, or to be fair add a condition to check skip index check unless both snapshots are provided.

modules/products/jira/locals.tf
nghazali: For Confluence I calculated the time from snapshot creation by adding another variable to define the snapshot creation date. I think this is the only way to be precise and don't make assumption.

nghazali: Having a long life cycle for index recovery files will prevent cleanup job to remove the older files, and if the environment is up and running for a some days we will have a new recovery index every day created by the scheduled job and eventually we run out of space in shared home. 

We should be wise to select this metric and as @errcode1202 suggested I think is good to make this configurable to be more precise in this area.

nghazali (changes requested):

modules/products/jira/locals.tf
jjeongatl: Shouldn't user always bring two in pair? When will be the case where user only wants to restore DB only?

nanux: I don't think we need to officially support restoring from the partial dataset. Yes, when we are creating those datasets, we will probably need to do something like this but that is a different use case than actually using the infrastructure.

nanux: But at the same time, I think it is possible that the shared home dataset might not contain the index snapshot. What would happen if we leave this in?

modules/products/jira/locals.tf
nanux: I don't think we need to be precise. We just need to convince Jira that it should pick up any snapshot that is available in the shared home. We don't get any benefit from people configuring this property. The index either is present or not, when it is put together with the database and shared home from the same timestamp, it should be consistent.

nanux, line 50: Can we extract `87600` into `locals.tf` with some descriptive name for the variable so it is not hidden here as a magic number?

nanux (approved):

modules/products/jira/locals.tf
```diff
@@ -16,6 +16,7 @@ locals {
 
   domain_supplied     = var.ingress.outputs.domain != null ? true : false
   product_domain_name = local.domain_supplied ? "${local.product_name}.${var.ingress.outputs.domain}" : null
+  ageOfUsableIndexSnapshot = 24 * 365 * 10
```
nanux, line 19:
```suggestion
  ageOfUsableIndexSnapshot = 24 * 365 * 10 # 10 years
```

PR #224, opened 2022-05-03
TFKUBE-454: Restore EBS snapshot from index recovery file

nghazali:
## Pull request description

Restore EBS snapshot from index recovery file with no need to full re-index. The snapshots for RDS and EBS should be match and EBS snapshot should contain index recovery files. 
Also user should pass the date of EBS snapshot creation. This will be used to calculate the age of the snapshot to extend the lifecycle to avoid out of date error (by default the max valid age of snapshot is 2 days). 

## Checklist


nghazali:
TODO: Documentation for the introduced variable (`confluence_shared_home_snapshot_creation_date`)

modules/products/confluence/locals.tf
```diff
@@ -87,4 +87,55 @@ locals {
...
+    }
+  }) : yamlencode({})
+
+  number_of_threads = min(4, floor(tonumber(local.confluence_software_resources.cpu)))
+
+  # Set the number of threads to boost re-index process based on number of Confluence CPUs per node
+  extend_reindex_thread_counts = yamlencode({
+    confluence = {
+      additionalJvmArgs = [
+        "-Dreindex.thread.count=${local.number_of_threads}",
```
nanux, line 134: I think we can keep these properties if they are generally improving indexing.

modules/products/confluence/locals.tf
```diff
@@ -87,4 +87,55 @@ locals {
...
+    confluence = {
+      additionalJvmArgs = [
+        "-Dcom.atlassian.confluence.journal.timeToLiveInMillis=${local.days_from_snapshot * local.day_millis}", # milliseconds
+        "-Dconfluence.cluster.index.recovery.generation.timeout=480", # seconds
+        "-Dconfluence.cluster.snapshot.file.wait.time=480", # seconds
+      ]
+    }
+  }) : yamlencode({})
+
+  number_of_threads = min(4, floor(tonumber(local.confluence_software_resources.cpu)))
```
nanux, line 128: This assumes we always get an integer (not `2500m` for example). The `confluence_cpu` is defined as a `string` type so it is possible to pass it in a different format. We can potentially make it more explicit, convert it here or at least document it.

modules/products/confluence/locals.tf
```diff
@@ -87,4 +87,55 @@ locals {
...
+  # Calculate the days passed from snapshot creation + 7 days extra
+  offset       = ((local.snapshot_month - 1 ) * 30) + local.snapshot_day
+  days_from_snapshot = (local.date_year - local.snapshot_year) * 365 + local.date_month * 30 + local.date_days - local.offset + 7
+
+  # Override the lifetime of recovery index in shared home and make sure there is enough timeout for copy long files
+  extend_snapshot_validity = var.db_snapshot_id != null ? yamlencode({
+    confluence = {
+      additionalJvmArgs = [
+        "-Dcom.atlassian.confluence.journal.timeToLiveInMillis=${local.days_from_snapshot * local.day_millis}", # milliseconds
+        "-Dconfluence.cluster.index.recovery.generation.timeout=480", # seconds
```
nanux, line 122: We can keep these properties, maybe extracting the `480` values into `locals.tf` to give them some documentation and sensible variable names.

nanux (changes requested):
I don't think the additional complexity and configurability are bringing a lot of customer value. We need to just convince Confluence that it should recover any index it finds in the shared home. Ideally if we would have some `confluence.validate.index.snapshot.timestamp = false` property. Unfortunatelly we don't have that so I think setting the value to some very high number as we are doing https://github.com/atlassian-labs/data-center-terraform/pull/223 is sufficient and easier to maintain.

yzha645:
As long as all of `main`, `change` and `edge` index are in place and in sync with the records in database, Confluence will be recovered from index snapshot stored in shared home. The age of snapshot is not a problem here, since index recovery is not actually checking it during startup.

In terms of the three properties regarding reindex thread counts, they either already have the default value 4, or already doing some calculation to dynamically decide the value based on workload and available cpus. Seems like we're re-enforcing Confluence features. https://confluence.atlassian.com/conf713/recognized-system-properties-1077914597.html 

Will close this PR for now, we can re-visit it later on if necessary.

PR #231, opened 2022-05-11
TFKUBE-461: Update license in shared home if restoring from snapshot

nanux:
## Pull request description

The license key is stored in the EBS volume with the shared home content when we are restoring from a snapshot. The license wasn't updated with the one used in the `config.tfvars`.

The change adds a command to replace the stored license with the value defined in the config file.

## Checklist


modules/products/confluence/locals.tf
```diff
@@ -87,4 +87,10 @@ locals {
       ]
     }
   }) : yamlencode({})
+
+  # updates base url (in case we are restoring from snapshot)
+  cmd_psql_update = "BASE_URL_TO_REPLACE=$(PGPASSWORD=${var.db_master_password} psql postgresql://${module.database.rds_endpoint}/${local.product_name} -U ${var.db_master_username} -Atc \"select BANDANAVALUE from BANDANA where BANDANACONTEXT = '_GLOBAL' and BANDANAKEY = 'atlassian.confluence.settings';\" | grep -i '<baseurl>'); PGPASSWORD=${var.db_master_password} psql postgresql://${module.database.rds_endpoint}/${local.product_name} -U ${var.db_master_username} -c \"update BANDANA set BANDANAVALUE = replace(BANDANAVALUE, '$${BASE_URL_TO_REPLACE}', '<baseUrl>${local.confluence_ingress_url}</baseUrl>') where BANDANACONTEXT = '_GLOBAL' and BANDANAKEY = 'atlassian.confluence.settings';\""
+
+  # updates license in shared home (in case we are restoring from snapshot)
+  cmd_license_update = "sed -i 's|<property name=\"atlassian.license.message\">.*</property>|<property name=\"atlassian.license.message\">${var.confluence_configuration["license"]}</property>|g' /shared-home/confluence.cfg.xml"
```
yzha645, line 95: I had a brief look at it a couple of days ago, and was thinking it's too much trouble to escape `/` in the license value. `|` is a better option. Good to know!

modules/products/confluence/pre_install.tf
```diff
@@ -12,9 +12,19 @@ resource "kubernetes_job" "pre_install" {
...
+          volume_mount {
+            mount_path = "/shared-home"
+            name       = "nfs-shared-home"
+          }
         }
         restart_policy = "Never"
+        volume {
+          name = "nfs-shared-home"
+          persistent_volume_claim {
+            claim_name = module.nfs.nfs_claim_name
```
yzha645, line 25: Do we need to depends_on module.nfs as well?

yzha645, line 25: Do we need to depends_on module.nfs as well?

nanux: I think the fact that it defines the variable from the module will inform Terraform that there is a dependency and it doesn't need to have `depends_on`

yzha645 (approved):

PR #243, opened 2022-07-25
CLIP-1609: Add `eks_additional_roles` variable

nanux:
## Pull request description

Enable the ability to access the cluster with a different role than the one used for cluster creation. For example if the cluster was created with CI, user can define additional (personal) role to access the cluster.

## Checklist


nanux:
Screen Shot 2022-07-25 at 9 32 58 pm

.github/workflows/e2e-test-no-domain.yaml
```diff
@@ -9,10 +9,13 @@ on:
       - 'docs/**'
       - '.atlassian/**'
       - 'test/unittest**'
+  pull_request_target:
```
nanux, line 12: Used the same technique to start `e2e` test when labeling a PR.

docs/docs/userguide/INSTALLATION.md
```diff
@@ -11,7 +11,7 @@ Set up a user with an administrator IAM role. See [Configuration basics — AWS
 Clone the `data-center-terraform` project repository from GitHub:
 
 ```shell
-git clone -b 2.0.0 https://github.com/atlassian-labs/data-center-terraform.git && cd data-center-terraform
+git clone -b 2.0.2 https://github.com/atlassian-labs/data-center-terraform.git && cd data-center-terraform
```
nanux, line 14: Updating to latest.

go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/atlassian-labs/data-center-terraform
 
-go 1.17
+go 1.18
 
 require (
 	github.com/aws/aws-sdk-go v1.41.16
 	github.com/go-git/go-git/v5 v5.4.2
-	github.com/gruntwork-io/terratest v0.38.8
+	github.com/gruntwork-io/terratest v0.40.18
```
nanux, line 8: This was causing alert for dependabot.

modules/AWS/eks/main.tf
```diff
@@ -39,4 +39,6 @@ module "eks" {
       capacity_type  = "ON_DEMAND"
     }
   }
+
+  map_roles = var.additional_roles
```
nanux, line 43: I looked into adding unit test for this but seems that the auth-aws value is not accessible in `terratest`.

config.tfvars
```diff
@@ -41,6 +41,18 @@ instance_disk_size = 50
...
+#eks_additional_roles = [
+#  {
+#    rolearn  = "arn:aws:iam::ACCOUNT_ID:role/ROLE_NAME"
+#    username = "ROLE_NAME"
+#    groups = [
+#      "system:masters"
+#    ]
+#  }
+#]
+
```
nghazali, line 55: We can remove the commented block if is not needed anymore.

yzha645: Do you think it's worth adding our `dcd-private-sysadmin` role to [e2e test config template](https://github.com/atlassian-labs/data-center-terraform/blob/main/test/e2etest/test-config.tfvars.tmpl), so that we're able to access cluster if e2e test failed?

yzha645 (approved):

config.tfvars
nanux: > We can remove the commented block if is not needed anymore.

The block there is an example (it has the `ACCOUNT_ID` and `ROLE_NAME` as placeholders)

nanux: > Do you think it's worth adding our `dcd-private-sysadmin` role to [e2e test config template](https://github.com/atlassian-labs/data-center-terraform/blob/main/test/e2etest/test-config.tfvars.tmpl), so that we're able to access cluster if e2e test failed?

That is a good point, I will create another PR for that.

PR #252, opened 2022-09-08
CLIP-1641: Resolve EKS Cycle Issue

jjeongatl:
## Pull request description

DCAPT team reported the terraform cycle error. This only happens on second run of install.sh as terraform tries to replace products PV due to dynamic provision of NFS service IP.
[
Screen Shot 2022-08-31 at 1 33 02 pm
](url)

This PR is to avoid the unnecessary replacement action by providing static IP address to every NFS servers' serviceIP.
Screen Shot 2022-09-08 at 1 30 30 pm

Also E2E test will run install.sh script twice to cover use case of 're-apply'.

## Checklist

### DIFF
--- a/.github/workflows/e2e-test-no-domain.yaml
+++ b/.github/workflows/e2e-test-no-domain.yaml
@@ -32,8 +32,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
+        run: echo ${GITHUB_REF}
 
       - name: Pin Kubectl version
         uses: azure/setup-kubectl@v2.0
--- a/.github/workflows/e2e-test-no-domain.yaml
+++ b/.github/workflows/e2e-test-no-domain.yaml
@@ -33,9 +33,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Print GITHUB_REF
-        run: echo ${GITHUB_REF}
-
       - name: Pin Kubectl version
         uses: azure/setup-kubectl@v2.0
         with:
--- a/.github/workflows/e2e-test-no-domain.yaml
+++ b/.github/workflows/e2e-test-no-domain.yaml
@@ -32,6 +32,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+
+      - name: Print GITHUB_REF
         run: echo ${GITHUB_REF}
 
       - name: Pin Kubectl version
--- a/.github/workflows/e2e-test-no-domain.yaml
+++ b/.github/workflows/e2e-test-no-domain.yaml
@@ -32,6 +32,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Pin Kubectl version
         uses: azure/setup-kubectl@v2.0
--- a/modules/AWS/eks/main.tf
+++ b/modules/AWS/eks/main.tf
@@ -12,8 +12,9 @@ module "eks" {
   enable_irsa = true
 
   # Networking
-  vpc_id  = var.vpc_id
-  subnets = var.subnets
+  vpc_id                    = var.vpc_id
+  subnets                   = var.subnets
+  cluster_service_ipv4_cidr = "172.20.0.0/16"
 
   # These 2 properties below will be deprecated in v18 of the AWS EKS module:
   # https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/UPGRADE-18.0.md
--- a/modules/AWS/eks/main.tf
+++ b/modules/AWS/eks/main.tf
@@ -14,7 +14,7 @@ module "eks" {
   # Networking
   vpc_id                    = var.vpc_id
   subnets                   = var.subnets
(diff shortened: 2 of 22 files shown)


modules/AWS/eks/main.tf
```diff
@@ -12,8 +12,9 @@ module "eks" {
   enable_irsa = true
 
   # Networking
-  vpc_id  = var.vpc_id
-  subnets = var.subnets
+  vpc_id                    = var.vpc_id
+  subnets                   = var.subnets
+  cluster_service_ipv4_cidr = "172.20.0.0/16"
```
jjeongatl, line 17: Restricting IPv4 range for every serviceIP in K8S cluster

modules/products/bamboo/main.tf
```diff
@@ -30,12 +30,13 @@ module "database" {
...
-  shared_home_size  = var.shared_home_size
+  namespace            = var.namespace
+  product              = local.product_name
+  requests_cpu         = var.nfs_requests_cpu
+  requests_memory      = var.nfs_requests_memory
+  limits_cpu           = var.nfs_limits_cpu
+  limits_memory        = var.nfs_limits_memory
+  availability_zone    = var.eks.availability_zone
+  shared_home_size     = var.shared_home_size
+  cluster_service_ipv4 = "172.20.2.2"
```
jjeongatl, line 41: lower band was picked to avoid collision
This is based on K8S IP address distribution algorithm
https://kubernetes.io/blog/2022/05/23/service-ip-dynamic-and-static-allocation/#avoid-ClusterIP-conflict

test/e2etest/installer_test.go
```diff
@@ -56,6 +57,10 @@ func runInstallScript(configPath string) {
 
 	// wait `cmd` until it finishes
 	_ = cmd.Wait()
+
+	// run again with same config. This is to assure the `re-apply` action works without any issue.
+	_ = cmd.Start()
+	_ = cmd.Wait()
```
jjeongatl, line 63: Re apply test

yzha645 (approved):

test/e2etest/installer_test.go
yzha645: Did it run for a second time? I couldn't find it in the Github action..

modules/products/bitbucket/pre_install.tf
```diff
@@ -1,4 +1,7 @@
 resource "kubernetes_job" "pre_install" {
+  lifecycle {
+    ignore_changes = all
+  }
```
bianchi2, line 4: `<>` symbols get into job command as:

```
'\u003cbaseUrl\u003ehttp://aabd8af13d1fc4d349258ef27546010c-1160533678.us-east-1.elb.amazonaws.com/confluence\u003c/baseUrl\u003e')
```
As a result, when running terraform apply for the second time, Terraform wants to delete the job and recreate it because pod spec is immutable and cannot be updated once the pod has been created.

Ignoring changes fixes the cycle error. Also, we don't expect to run this job twice i.e. restore environment from a different snapshot if it has been already restored from a snapshot.

jjeongatl: It is an acceptable trade off as it is less likely to restore multiple dataset snapshots on one environment.

bianchi2:
LGTM

.github/workflows/e2e-test-no-domain.yaml
```diff
@@ -33,6 +33,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Print GITHUB_REF
```
nghazali, line 36: How this would be useful?

modules/AWS/eks/main.tf
nghazali: Please add this in local to avoid the literal in code.

modules/products/bamboo/main.tf
nghazali: again, please use local variables rather than the constant literals.

modules/products/confluence/main.tf
```diff
@@ -43,4 +43,5 @@ module "nfs" {
   availability_zone       = var.eks.availability_zone
   shared_home_snapshot_id = var.shared_home_snapshot_id
   shared_home_size        = var.shared_home_size
+  cluster_service_ipv4    = "172.20.2.4"
```
nghazali, line 46: The same here

modules/products/jira/main.tf
```diff
@@ -25,6 +25,8 @@ module "nfs" {
   availability_zone       = var.eks.availability_zone
   shared_home_snapshot_id = var.shared_home_snapshot_id
   shared_home_size        = var.shared_home_size
+  cluster_service_ipv4    = "172.20.2.5"
```
nghazali, line 28: And this. As there are multiple constant for service IP, we may better to centralised them.

test/e2etest/installer_test.go
jjeongatl: I just confirmed with manual run and it worked

nghazali:
After test the following scenario and pass e2e tests, feel free to merge the changes:

- Install at least one product using terraform solution
- Run `install.sh` with the same configuration
- Confirm installation completed with no error and detect no config change
- Add one more node (replica_count) and run install again
- Confirm the scaling completed successfully
- Uninstall the infrastructure and confirm there is no error

nghazali (approved):

bianchi2 (approved):

PR #253, opened 2022-09-26
CLIP-1656: Override termination graceful period

bianchi2:
## Pull request description

Under certain conditions application pods are stuck in terminating when a Helm chart is deleted which blocks deletion of a shared home PVC. This results in a timeout error when running uninstall.sh. This PR makes termination grace period configurable in config var file which is passed to helm resource when it's being created. 

No unit tests since this part of pod spec is hidden in kube runtime which isn't available to Terraform.

## Checklist

### DIFF
--- a/modules/products/bamboo/variables.tf
+++ b/modules/products/bamboo/variables.tf
@@ -50,7 +50,7 @@ variable "termination_grace_period" {
   description = "Termination grace period in seconds"
   type        = number
   validation {
-    condition     = var.termination_grace_period > 0
+    condition     = var.termination_grace_period >= 0
     error_message = "Termination grace period needs to be a positive number."
   }
 }
--- a/modules/products/bamboo/variables.tf
+++ b/modules/products/bamboo/variables.tf
@@ -49,7 +49,10 @@ variable "installation_timeout" {
 variable "termination_grace_period" {
   description = "Termination grace period in seconds"
   type        = number
-  default     = 30
+  validation {
+    condition     = var.termination_grace_period > 0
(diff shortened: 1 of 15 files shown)


modules/products/bamboo/variables.tf
```diff
@@ -46,6 +46,12 @@ variable "installation_timeout" {
   }
 }
 
+variable "termination_grace_period" {
+  description = "Termination grace period in seconds"
+  type        = number
+  default     = 30
```
nghazali, line 52: As we have default in root level, we can remove the default (make it mandatory for module) here and instead we may add a validator to limit the grace period to a logical range. WDYT?

The same comment for this variable in other modules.

bianchi2: Yeah, I was thinking about it. Let me do that!

modules/products/bamboo/variables.tf
```diff
@@ -49,7 +49,10 @@ variable "installation_timeout" {
 variable "termination_grace_period" {
   description = "Termination grace period in seconds"
   type        = number
-  default     = 30
+  validation {
+    condition     = var.termination_grace_period > 0
```
nghazali, line 53: Can't the grace period be `0`? 
I suggest to set a range to something like [0, 5 minutes]

bianchi2: The very purpose of this change is to be able to make it 0 :) Let me fix it

nghazali (approved):

nghazali:
We have failed unittest!

jjeongatl (approved):

PR #260, opened 2022-10-05
CLIP-1669: Make it possible to disable 443 port for nginx ingress

bianchi2:
This PR does not change any default behavior but rather introduces a new variable `enable_https_ingress`. Sometimes, Nginx ingress helm chart can't be deployed because a loadbalancer can't be created for svc. This happens when a lot of cidrs are provided in `whitelist_cidr`, as a result the controller tries to create too many inbound rules in the security group and fails to do so because of the limit:

```
  Warning  SyncLoadBalancerFailed  112s                service-controller  Error syncing load balancer: failed to ensure load balancer: error authorizing security group ingress: "RulesPerSecurityGroupLimitExceeded: The maximum number of rules per security group has been reached.\n\tstatus code: 400, request id: 7de945ea-0571-48cd-99a1-c2ca528a441c"
```
Since rules are created for 80 and 443, it is possible to hit this limit. This PR introduces a way to enable only 80 port, i.e. the LB won't listen on 443 (which is an OK case in most CI related scenarios)

## Checklist


variables.tf
```diff
@@ -118,6 +118,12 @@ variable "whitelist_cidr" {
   }
 }
 
+variable "enable_https_ingress" {
+  description = "If true, Nginx controller will listen on both 80 and 443."
```
nghazali, line 122: As this variable has nothing with port 80, so this description is not clear what is the behaviour if the value is False. 
We may say `If true, Nginx controller will listen to 443 as well.`.

nghazali:
Is there any way to increase the security group limit instead? I'm kind of feeling this is sort of hack solution. Imagine if in a use case, customer needs to have both 80 and 443, but hit this issue! In such scenario this solution is not helping.

bianchi2:
Yes, this is a workaround (which although may help in cases when 30 cidrs are provided). And yes, it is possible to request quota increase. With this PR merged though, one can cut the number of rules in half though (if 443 isn't required)

variables.tf
bianchi2: Description updated

nghazali:
> Yes, this is a workaround (which although may help in cases when 30 cidrs are provided). And yes, it is possible to request quota increase. With this PR merged though, one can cut the number of rules in half though (if 443 isn't required), which can help even with increased quotas provided that the number of whitelisted cidrs is huge.

Thanks, it makes sense. I think is good to add a use case in troubleshooting page, and provide the suggestion/workaround to inform users what to do if the problem happened. I kind of feeling we add this feature, but is not really visible if the problem happened and users will raise a support ticket and we should reply them case by case. At least having this case in troubleshooting helps to reduce the number of support tickets in future. WDYT?

bianchi2:
Agreed, I'll add an abstract or two in troubleshooting docs with exact details on what causes the issue and how to mitigate it (including dropping one port)

config.tfvars
```diff
@@ -23,6 +23,11 @@ products = ["<LIST_OF_PRODUCTS>"]
 # default value to your desired CIDR blocks. e.g. ["10.20.0.0/16" , "99.68.64.0/10"]
 whitelist_cidr = ["0.0.0.0/0"]
 
+# Only enable http port 80 with Nginx ingress controller. By default, Ingress controller listens on 443 and 80.
```
nanux, line 26:
```
# By default, Ingress controller listens on 443 and 80. You can enable only http port 80 by 
# uncommenting the below line, which will disable port 443. This results in fewer inbound rules in Nginx controller security group.
# This can be used in case you hit the limit which can happen if 30+ whitelist_cidrs are provided.
#enable_https_ingress = false
```

nanux (approved):

nanux:
Please add a short explanation into https://github.com/atlassian-labs/data-center-terraform/blob/main/docs/docs/troubleshooting/TROUBLESHOOTING.md

nghazali (approved):

PR #262, opened 2022-10-07
Fix synchrony test

bianchi2:
Currently, synchrony tests are failing https://github.com/atlassian-labs/data-center-terraform/actions/runs/3201163413/jobs/5228854594

This happens because tests use the wrong synchrony url - it has /confluence in it while it should not https://github.com/atlassian/data-center-helm-charts/blob/main/src/main/charts/confluence/templates/ingress.yaml#L40

This PR trims `/confluence` suffix

## Checklist

### DIFF
--- a/modules/products/confluence/locals.tf
+++ b/modules/products/confluence/locals.tf
@@ -48,12 +48,11 @@ locals {
 
   confluence_ingress_url = local.domain_supplied ? "https://${local.product_domain_name}" : "http://${var.ingress.outputs.lb_hostname}/${local.product_name}"
 
-  synchrony_ingress_url = local.domain_supplied ? "${local.confluence_ingress_url}/synchrony" : "http://${var.ingress.outputs.lb_hostname}/${local.product_name}/synchrony"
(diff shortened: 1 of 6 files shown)


modules/products/confluence/locals.tf
```diff
@@ -48,12 +48,11 @@ locals {
 
   confluence_ingress_url = local.domain_supplied ? "https://${local.product_domain_name}" : "http://${var.ingress.outputs.lb_hostname}/${local.product_name}"
 
-  synchrony_ingress_url = local.domain_supplied ? "${local.confluence_ingress_url}/synchrony" : "http://${var.ingress.outputs.lb_hostname}/${local.product_name}/synchrony"
+  synchrony_ingress_url = local.domain_supplied ? "${local.confluence_ingress_url}/synchrony" : "http://${var.ingress.outputs.lb_hostname}/synchrony"
 
   synchrony_settings_stanza = yamlencode({
     synchrony = {
       enabled    = true
-      ingressUrl = local.synchrony_ingress_url
```
bianchi2, line 56: No ingressUrl for synchrony in values.yaml anymore

yzha645 (approved):

PR #266, opened 2022-10-17
CLIP-1677: Introduced the configuration for Synchrony instance

nghazali:
## Pull request description

Add configuration to customise the synchrony instance:
- CPU
- Memory
- JVM max/min heap size and stack size

## Checklist

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -377,6 +377,36 @@ variable "confluence_max_heap" {
   default     = "512m"
 }
 
+variable "synchrony_cpu" {
+  description = "Number of CPUs for synchrony instance"
+  type        = string
+  default     = "2"
+}
+
+variable "synchrony_mem" {
+  description = "Amount of memory for synchrony instance"
+  type        = string
+  default     = "2.5Gi"
+}
+
+variable "synchrony_min_heap" {
+  description = "Minimum heap size for synchrony instance"
+  type        = string
+  default     = "1Gi"
+}
+
+variable "synchrony_max_heap" {
+  description = "Maximum heap size for synchrony instance"
+  type        = string
+  default     = "2Gi"
+}
+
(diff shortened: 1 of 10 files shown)


bianchi2 (approved):

variables.tf
```diff
@@ -377,6 +377,36 @@ variable "confluence_max_heap" {
...
+variable "synchrony_mem" {
+  description = "Amount of memory for synchrony instance"
+  type        = string
+  default     = "2.5Gi"
+}
+
+variable "synchrony_min_heap" {
+  description = "Minimum heap size for synchrony instance"
+  type        = string
+  default     = "1Gi"
```
bianchi2, line 395: It has be be `1g` otherwise JVM won't be created - https://github.com/atlassian-labs/data-center-terraform/actions/runs/3262312392/jobs/5359248528#step:11:10025

These are logs from synchrony pod:
```
Invalid initial heap size: -Xms1Gi
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

variables.tf
```diff
@@ -377,6 +377,36 @@ variable "confluence_max_heap" {
...
+}
+
+variable "synchrony_max_heap" {
+  description = "Maximum heap size for synchrony instance"
+  type        = string
+  default     = "2Gi"
```
bianchi2, line 401: Same here

bianchi2 (changes requested):

variables.tf
bianchi2, line 395:
```suggestion
  default     = "1g"
```

variables.tf
bianchi2, line 401:
```suggestion
  default     = "2g"
```

bianchi2 (approved):

PR #270, opened 2022-10-21
Pin Python version in setup python workflow step

bianchi2:
Since we export [PYTHONPATH in cmd.exec](https://github.com/atlassian-labs/data-center-terraform/blob/main/test/e2etest/installer_test.go#L114), Python version should be the same to be able to run aws_clean.py that uses boto3 library.


bianchi2:
Merging with no approvals to avoid failed builds

PR #273, opened 2022-10-24
CLIP-1646: Update aws and rds providers

bianchi2:
This PR updates aws provider version to 4.36 and rds provider version to 5.1. Both are the latest as of now.

The following attributes were added to rds module (because they are true by default):

```
storage_encrypted          = false
create_random_password     = false
create_db_subnet_group      = true
```

Also, the following was changed in outputs:

```
output "rds_master_password" {
  value     = module.db.db_instance_password
  sensitive = true
}
```

No changes to any other aws resources.

This fixes a warning on each and every tf apply and destroy

```
Warning: Redundant ignore_changes element  on .terraform/modules/bamboo.database.db/modules/db_instance/main.tf line 20, in resource "aws_db_instance" "this":
  20: resource "aws_db_instance" "this" {
```

Important! Run `terraform init --update` to fetch new provider versions. This change is backward compatible, i.e. can be applied to infra created before version update. This still has to be extensively tested, that's why it's safer to use it to create new environments only.

## Checklist

### DIFF
--- a/modules/AWS/rds/main.tf
+++ b/modules/AWS/rds/main.tf
@@ -42,8 +42,8 @@ module "db" {
 
   db_name                     = var.db_name
   username                    = local.db_master_username
-  password                    = local.db_master_password
-  create_random_password      = false
+  password                    = var.db_master_password
+  create_random_password      = local.create_random_password
   port                        = 5432
 
   create_db_subnet_group      = true
@@ -53,6 +53,7 @@ module "db" {
   maintenance_window          = "Mon:00:00-Mon:03:00"
   backup_window               = "03:00-06:00"
   storage_encrypted           = false
+
   # Snapshot settings
   snapshot_identifier         = var.snapshot_identifier
   allow_major_version_upgrade = var.snapshot_identifier != null
@@ -61,10 +62,4 @@ module "db" {
 
   skip_final_snapshot         = true
   apply_immediately           = true
-}
-
-resource "random_password" "password" {
-  length           = 12
-  special          = true
-  override_special = "!#$%^*(){}?,."
 }
--- a/modules/AWS/rds/main.tf
+++ b/modules/AWS/rds/main.tf
@@ -22,43 +22,44 @@ module "security_group" {
 }
 
 module "db" {
-  source  = "terraform-aws-modules/rds/aws"
-  version = "~> 3.0"
+  source                      = "terraform-aws-modules/rds/aws"
+  version                     = "~> 5.1"
 
-  identifier = var.rds_instance_identifier
+  identifier                  = var.rds_instance_identifier
 
-  create_db_option_group    = false
-  create_db_parameter_group = false
+  create_db_option_group      = false
+  create_db_parameter_group   = false
 
   # All available versions: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts
-  engine               = "postgres"
-  engine_version       = local.engine_version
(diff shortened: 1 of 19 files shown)


tarka (approved):

modules/AWS/rds/main.tf
```diff
@@ -22,47 +22,49 @@ module "security_group" {
...
+  iops                        = var.iops
 
-  name     = var.db_name
-  username = local.db_master_username
-  password = local.db_master_password
-  port     = 5432
+  db_name                     = var.db_name
+  username                    = local.db_master_username
+  password                    = local.db_master_password
+  create_random_password      = false
```
nghazali, line 46: It doesn't harm to set `create_random_password` to `false`, however based on the implementation, it never being used and the value of this property is not important because `password` variable always has a value in this module.

modules/AWS/rds/main.tf
```diff
@@ -22,47 +22,49 @@ module "security_group" {
 }
 
 module "db" {
-  source  = "terraform-aws-modules/rds/aws"
-  version = "~> 3.0"
+  source                      = "terraform-aws-modules/rds/aws"
```
nghazali, line 25: I wonder why the indentation of this file has changed? Based on terraform linting, this kind of change should raise an issue before commit and I suspect you have not the `pre-commit` hook installed properly on your local machine. (see [pre-commit hook](https://atlassian-labs.github.io/data-center-terraform/development/HOW_TO_START/))

nghazali (changes requested):

modules/AWS/rds/main.tf
jjeongatl, line 46: If we are going to change it to false we need to update comments in config.tfvars.
https://github.com/atlassian-labs/data-center-terraform/blob/ef2d9920f88215a57597d937e6ab5ec82caf1eb5/config.tfvars#L174

jjeongatl (changes requested):

modules/AWS/rds/main.tf
bianchi2: @nghazali in the latest version of this module `create_random_password` is set to True by default, and password is ignored in this case:

> Description: Password for the master DB user. Note that this may show up in logs, and it will be stored in the state file. The password provided will not be used if the variable create_random_password is set to true.

bianchi2: We will still generate and use password. This attribute is for the module itself to generate one

bianchi2: I am going to rework it and get rid of password generation to let rds module do that for us if password isn't set.

modules/AWS/rds/main.tf
bianchi2: I guess I changed in deliberately to have a table like representation of key-values.

test/unittest/database_test.go
```diff
@@ -66,12 +66,6 @@ func TestDbVariablesPopulatedWithValidValues(t *testing.T) {
 	assert.EqualValues(t, inputAllocatedStorage, planAllocatedStorage)
 	assert.EqualValues(t, inputIops, planIops)
 	assert.EqualValues(t, true, planApplyImmediately)
-
```
bianchi2, line 69: Now that we don't generate password, we rely on rds module to do it from us

jjeongatl (approved):

modules/AWS/rds/main.tf
nghazali: Thanks, make sense

nghazali (approved):

PR #275, opened 2022-10-30
CLIP-1699: Update versions for 2.2.0 release.

tarka:
## Pull request description

Minor release as it contains a slight behavioural change (bidirectional PV)

## Checklist
- [/] I have user documentation (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Change Log
 
-## 2.2.0
+## 2.1.1
 
-**Release date:** 2022-10-31
+**Release date:** 2022-11-01
 
 * PV binding are now bidirectional
 * Update and/or pin some dependency versions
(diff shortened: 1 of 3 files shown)


bianchi2 (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log
 
+## 2.2.0
```
nghazali, line 3: This is a bug fix release, the release number should be `2.1.1`

nghazali (changes requested):

CHANGELOG.md
tarka: The bidirectional PV feels more like a minor functional change. @bianchi2 what do you think?

nghazali: Still this is a bug fix. This fixes the problem of not removing pv in uninstall in some cases.

bianchi2: It's just a smarter way to create volumes - we make sure that the volume we create for a DC app does not get bound to elastic search (or any other 3rd party claim)

PR #277, opened 2022-10-31
CLIP-1570: Update eks module version to 18.30.2

bianchi2:
Major refactoring as per https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/docs/UPGRADE-18.0.md

The biggest changes are:

* node_groups has become `eks_managed_node_groups`
* IAM role for worker nodes is pre-created to be able to timely attach required osquery policies
* a few attributes have been removed or renamed
* no local.file with kubeconfig anymore (it is generated in the install script) - also context name has changed to full role arn

Autoscaling group keeps using a custom launch template even though in a new module it is possible to provide pre_boostrap user_data. However, the problem with tagging EC2 instances comes back. Thus, no changes in how a custom user_data in a launch template is used

## Checklist


install.sh
```diff
@@ -206,6 +206,8 @@ set_current_context_k8s() {
   local EKS_CLUSTER="${EKS_CLUSTER_NAME:0:38}"
   CONTEXT_FILE="${ROOT_PATH}/kubeconfig_${EKS_CLUSTER}"
 
+  aws eks update-kubeconfig --name "${EKS_CLUSTER}" --region "${REGION}" --kubeconfig ${CONTEXT_FILE}
```
bianchi2, line 209: We need to generate kubeconfig to a custom location where e2e find it

modules/AWS/eks/iam.tf
```diff
@@ -0,0 +1,37 @@
+resource "aws_iam_role" "node_group" {
```
bianchi2, line 1: Create iam role and attach EKS worker node policies to it + osquery policies (if enabled). Due to a bug, it's not possible to use `iam_role_additional_policies` and let the module automatically create iam role. Such policies need to be defined (not computed) before creating the nodegroup.

modules/AWS/eks/locals.tf
```diff
@@ -13,9 +13,10 @@ locals {
 
   cluster_service_ipv4_cidr = "172.20.0.0/16"
 
-  workers_additional_policies = var.osquery_secret_name != "" ? [aws_iam_policy.laas[0].arn,aws_iam_policy.fleet_enrollment_secret[0].arn] : []
-
   osquery_secret_region = var.osquery_secret_region != "" ? var.osquery_secret_region : var.region
 
   account_id = data.aws_caller_identity.current.account_id
+
+  eks_node_policies = ["arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore", "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy", "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly", "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy"]
```
bianchi2, line 20: If necessary, we can make it configurable in tfvars to let pass any custom policies to be attached to nodegroup iam role

modules/AWS/eks/main.tf
```diff
@@ -14,58 +15,65 @@ module "nodegroup_launch_template" {
...
+  version = "18.30.2"
 
   # Configure cluster
-  cluster_version              = "1.21"
-  cluster_name                 = var.cluster_name
-  manage_cluster_iam_resources = true
+  cluster_version = "1.21"
+  cluster_name    = var.cluster_name
+  create_iam_role = true
+  create_cloudwatch_log_group = false
```
bianchi2, line 24: e2e iam user does not have privileges to create log groups. In future, we can make it configurable as well (if eks logging is a requirement)

modules/AWS/eks/main.tf
```diff
@@ -14,58 +15,65 @@ module "nodegroup_launch_template" {
...
-      subnets                 = slice(var.subnets, 0, 1)
-      instance_types          = var.instance_types
-      capacity_type           = "ON_DEMAND"
+      name                         = "appNode-${replace(join("-", var.instance_types), ".", "_")}"
+      max_size                     = var.max_cluster_capacity
+      desired_size                 = var.min_cluster_capacity
+      min_size                     = var.min_cluster_capacity
+      subnet_ids                   = slice(var.subnets, 0, 1)
+      capacity_type                = "ON_DEMAND"
+      create_launch_template       = false
```
bianchi2, line 64: We use a custom launch template that is merged with the default one (which executes bootstrap script for the node to join the cluster)

modules/AWS/eks/outputs.tf
```diff
@@ -10,10 +10,6 @@ output "cluster_oidc_issuer_url" {
   value = module.eks.cluster_oidc_issuer_url
 }
 
-output "kubeconfig_filename" {
```
bianchi2, line 13: Removed in 18.x. Kubeconfig is generated in install.sh

modules/AWS/eks/outputs.tf
```diff
@@ -31,14 +27,6 @@ output "cluster_asg_name" {
   value = local.cluster_asg_name
 }
 
-output "max_cluster_size" {
```
bianchi2, line 34: Unused and deprecated

modules/AWS/eks/nodegroup_launch_template/locals.tf
```diff
@@ -25,6 +25,7 @@ locals {
   templates_all = var.osquery_secret_name != "" ? concat(tolist(local.templates), ["osquery/osquery.sh.tpl"]) : local.templates
 
   user_content              = [ for tpl in local.templates_all : templatefile("${path.module}/templates/${tpl}", {
+    cluster_name                    = var.cluster_name
```
nanux, line 28: Is this a new required value?

modules/AWS/eks/locals.tf
nanux: /nit we can break it into new lines

I think the configurabitlity is a good idea but let's wait when we know it is useful. It could be good to raise this as an improvement for Terraform.

modules/AWS/eks/nodegroup_launch_template/locals.tf
bianchi2: Thanks for catching that! It's unused indeed.

modules/AWS/eks/locals.tf
bianchi2: Yes, multi line list looks better. Fixed

modules/AWS/eks/iam.tf
tarka: Should these comments be in the file rather than the PR?

bianchi2: I actually added a few lines here https://github.com/atlassian-labs/data-center-terraform/blob/eec1d1f8bc8f1397ea2c003b661eba2a7cd83c3a/modules/AWS/eks/iam.tf#L25

nghazali (approved):

PR #278, opened 2022-11-02
CLIP-1570: Make iam role name more unique

bianchi2:
Just a small fixup to https://github.com/atlassian-labs/data-center-terraform/pull/277

It's better to have a unique iam role name to avoid conflicts if for some reason the role isn't deleted.

## Checklist


modules/AWS/eks/main.tf
```diff
@@ -39,6 +39,7 @@ module "eks" {
   # Enables IAM roles for service accounts - required for autoscaler and potentially Atlassian apps
   # https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
   enable_irsa = true
+  iam_role_use_name_prefix = false
```
bianchi2, line 42: To avoid hitting a 38 character limit for a prefix that is used in some resources that the module creates

jjeongatl (approved):

PR #279, opened 2022-11-02
CLIP-1707: Set termination grace period to 0 by default

bianchi2:
This PR sets `termination_grace_period` to 0 by default to avoid having pods stuck in Pending and holding deletion of PVC.

Also, removing overrides from test tfvars, and tweaking variable descrition

## Checklist

### DIFF
--- a/config.tfvars
+++ b/config.tfvars
@@ -110,7 +110,7 @@ jira_replica_count = 1
 # Termination grace period
 # Under certain conditions, pods may be stuck in a Terminating state which forces shared-home pvc to be stuck
 # in Terminating too causing Terraform destroy error (timing out waiting for a deleted PVC).
-# Termination grace period is 0 by default. You can override it if for some reason you need a different value
+# Termination grace period is 0 by default which instructs kubelet to forcefully terminate the pod. You can override it if for some reason you need a different value
 #jira_termination_grace_period = 0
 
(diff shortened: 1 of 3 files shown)


config.tfvars
```diff
@@ -109,8 +109,8 @@ jira_replica_count = 1
 
 # Termination grace period
 # Under certain conditions, pods may be stuck in a Terminating state which forces shared-home pvc to be stuck
-# in Terminating too causing Terraform destroy error (timing out waiting for a deleted PVC). Set termination graceful period to 0
-# if you encounter such an issue
+# in Terminating too causing Terraform destroy error (timing out waiting for a deleted PVC).
+# Termination grace period is 0 by default. You can override it if for some reason you need a different value
```
jjeongatl, line 113: Can we explain what does the variable do to prevent described issue? setting this value to zero means the pod will be forced shut down in 0 mins?

bianchi2: Setting to 0 is to instruct kubelet to forcefully delete the pod. More info https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/

nanux (approved):

nanux:
Are we experiencing the same issues with the standard helm chart installation? If the application is destroyed forcefully, it might not have flushed the state and finished some long-running processes.

bianchi2:
Closing in favor of https://github.com/atlassian-labs/data-center-terraform/pull/282

PR #283, opened 2022-11-14
CLIP-1673: Test with domain install 2 times in a row

bianchi2:
The issue with terraform apply 2 times in a row cannot be reproduced anymore. See:

* https://github.com/atlassian-labs/data-center-terraform/actions/runs/3457214411/jobs/5770494962#step:11:5251
* https://github.com/atlassian-labs/data-center-terraform/actions/runs/3457914723/jobs/5771815875#step:11:5261

Perhaps, updating AWS provider version has fixed it. 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ outputs.json
 
 # go packages
 test/pkg
+
+# File created by tagging resource
+modules/AWS/asg_ec2_tagging
--- a/test/e2etest/installer_test.go
+++ b/test/e2etest/installer_test.go
@@ -25,9 +25,7 @@ func TestInstaller(t *testing.T) {
 	runInstallScript(testConfig.ConfigPath)
 
 	// run again with same config. This is to assure that the `re-apply` action finishes without any issue.
-	if !useDomain {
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -65,4 +65,7 @@ docs/site
 outputs.json
 
 # go packages
-test/pkg
\ No newline at end of file
+test/pkg
+
+# File created by tagging resource
+modules/AWS/asg_ec2_tagging
```
bianchi2, line 71: aws tagging creates such a file, and it'd be nice to have it in gitignore to avoid committing it unintentionally.

yzha645: Do we still need tagging module while using launch template tags for eks?

yzha645 (approved):

bianchi2:
@yzha645 no, we don't need this module anymore. `modules/AWS/asg_ec2_tagging` is a file that TF creates after we have added  https://github.com/atlassian-labs/data-center-terraform/blob/main/modules/AWS/eks/main.tf#L75

PR #284, opened 2022-11-15
CLIP-1708: Fix use of local helm charts

bianchi2:
This PR adds support of installing local Helm charts for Jira and Bitbucket (smth that is there for Confluence and Bamboo), as well as unsets defaults for Helm chart version. When Helm chart versions are explicitly set they need to be constantly updated. When no version is provided, helm_release resource will automatically use the latest. It'll still be possible to override helm chart version, i.e. existing behavior is unchanged.

## Checklist


test/unittest/bamboo_test.go
```diff
@@ -76,7 +77,7 @@ var BambooCorrectVariables = map[string]interface{}{
 	"dataset_url":          nil,
 	"installation_timeout": testTimeout,
 	"bamboo_configuration": map[string]interface{}{
-		"helm_version": "1.0.0",
+		"helm_version": "",
```
bianchi2, line 80: Pass no version to make sure that in the helm_release resource the version is not null (Helm provider will automatically get the latest)

nghazali (approved):

PR #288, opened 2022-11-24
CLIP-1722: Set EKS version to 1.24 and make it configurable

bianchi2:
1.24 is currently [the latest supported EKS version](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html). This PR makes EKS version a configurable variable that defaults to 1.24. Also, ebs-csi-driver addon is added with the required policy for worker IAM role (See: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-1.23 on why this addon is added).

`eks_version` can be overridden and is validated by Terraform. A small unit test is added to check that.

## Checklist


modules/AWS/eks/locals.tf
```diff
@@ -20,5 +20,6 @@ locals {
   eks_node_policies = [ "arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore",
                         "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
                         "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
-                        "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy" ]
+                        "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+                        "arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy" ]
```
bianchi2, line 24: This one is required for ebs csi driver to attach EBS vols

errcode1202: Was this always missing then, just wondering how it would have ever worked?

modules/AWS/eks/variables.tf
```diff
@@ -7,6 +7,15 @@ variable "cluster_name" {
   }
 }
 
+variable "eks_version" {
+  description   = "EKS K8s version"
+  type          = number
+  validation {
+    condition     = can(regex("^1\\.2[1-4]", var.eks_version))
```
errcode1202, line 14: So we're enforcing only these versions `1.21, 1.22, 1.23, 1.24` because anything else is not tested?

errcode1202 (approved):

modules/AWS/eks/locals.tf
bianchi2: Previously, ebs attachment was done by the apiserver/controller (which has the powers), while now it's a separately deployed service that runs on the node.

modules/AWS/eks/variables.tf
bianchi2: Yes, EKS 1.21-1.24 is what's currently supported when you create [a new EKS cluster](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html).

PR #293, opened 2022-12-12
CLIP-1734: Make skip refresh optional

bianchi2:
While https://github.com/atlassian-labs/data-center-terraform/pull/291 fixes the error when destroying terraform after a failed terraform apply, it introduces another issue. Skipping refresh can sometimes make Helm provider keep using an invalidated token, which results in the following error:

```
Error: Kubernetes cluster unreachable: the server has asked for the client to provide credentials
```
This PR makes `-refresh=false` optional, and it can be added to terraform destroy args with `-s` flag. 2 additional FAQs were added as well.

## Checklist

### DIFF
--- a/docs/docs/troubleshooting/TROUBLESHOOTING.md
+++ b/docs/docs/troubleshooting/TROUBLESHOOTING.md
@@ -280,6 +280,7 @@ This guide contains general tips on how to investigate an application deployment
     Wait for the ASG to be deleted, then go back to EKS cluster > Delete.
 
 ??? tip "How to deal with `This object does not have an attribute named` error when running uninstall.sh"
+
     It is possible that if the installation has failed, the uninstall script will return an error like:
     ```
     module.base-infrastructure.module.eks.aws_autoscaling_group_tag.this["Name"]: Refreshing state... [id=eks-appNode-t3_xlarge-50c26268-ea57-5aee-4523-68f33af7dd71,Name]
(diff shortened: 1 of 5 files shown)


docs/docs/troubleshooting/TROUBLESHOOTING.md
```diff
@@ -2,15 +2,15 @@
 
 This guide contains general tips on how to investigate an application deployment that doesn't work correctly.
 
-??? tip "How do I uninstall an environment using a different Terraform configuration file?" 
-
```
nghazali, line 6: There are lot's of unnecessary changes comes from your editor settings which makes the PR very noisy, and because all other have different setting, probably if anyone else makes another change in the files they will repeat this noise again. I suggest change the rules on your editor to avoid those changes. I see yours end trims all spaces which means change in many lines.

bianchi2: Fixed by editing in GitHub. Somehow Atom would trim those spaces

nghazali (approved):

uninstall.sh
```diff
@@ -118,7 +124,7 @@ destroy_infrastructure() {
     terraform -chdir="${ROOT_PATH}" init -no-color | tee -a "${LOG_FILE}"
   fi
   set +e
-  terraform -chdir="${ROOT_PATH}" destroy -auto-approve -refresh=false -no-color "${OVERRIDE_CONFIG_FILE}" | tee -a "${LOG_FILE}"
+  terraform -chdir="${ROOT_PATH}" destroy -auto-approve ${SKIP_REFRESH} -no-color "${OVERRIDE_CONFIG_FILE}" | tee -a "${LOG_FILE}"
```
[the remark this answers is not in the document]
jjeongatl: for the sake of consistency

PR #297, opened 2022-12-21
CLIP-1749, CLIP-1744: Read engine_version from snapshot. Support Postgres 14

bianchi2:
This PR forces Terraform to rely on engine_version from a db snapshot if defined rather than on supplied values in tfvars file. This will fix the issue when snapshot db version differs from the one in tfvars.

Tested locally by using a snapshot with 11.4 version and setting 13.4 in tfvars. The first apply was a success and 11.4 db was created, the second one said there were no changes, i.e. engine version values from tfvars are ignored if db snapshot is defined.

Writing a unit test is somewhat problematic since the actual version is read from a real snapshot. To the contrary, I had to remove db snapshot identifier from unit tests as I found no way to avoid reading data and using some mock that would make Terratest happy. However, the logic around supplying db snapshot hasn't been changed so it's safe not to check if snapshot identifier is in db attributes.

Also, all DC apps support Postgres 14, so it's allowed in variable validation block, and major version 14 is mapped to 14.5 which is the latest 14.x as of today.

## Checklist

### DIFF
--- a/modules/AWS/rds/variables.tf
+++ b/modules/AWS/rds/variables.tf
@@ -48,7 +48,7 @@ variable "vpc" {
 }
 
 variable "major_engine_version" {
-  description = "RDS Major engine version for the product."
+  description = "RDS Major engine version for the product. If RDS snapshot is provided, this value is ignored."
   default     = "11"
   type        = string
   validation {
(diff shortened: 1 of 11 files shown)


modules/AWS/rds/variables.tf
```diff
@@ -52,8 +52,8 @@ variable "major_engine_version" {
   default     = "11"
```
nanux, line 52: Should we say something like:
```
description = "RDS Major engine version for the product. If DB snapshot is provided, this value is ignored."
```

nanux:
Nice improvement!

nanux (approved):

PR #298, opened 2022-12-22
CLIP-17-52: Add NAT IP to whitelist_cidr

bianchi2:
This PR fixes an issue with Confluence being unable to reach Synchrony using ingress URL when `whitelist_cidr` restricts access to a particular range. Since private subnets use NAT, we need to silently add NAT EIP to whitelist_cird, unless it's all-permissive `0.0.0.0/0`

## Checklist

### DIFF
--- a/modules/common/locals.tf
+++ b/modules/common/locals.tf
@@ -2,4 +2,5 @@ locals {
   vpc_name       = format("atlas-%s-vpc", var.environment_name)
   cluster_name   = format("atlas-%s-cluster", var.environment_name)
   ingress_domain = var.domain != null ? "${var.environment_name}.${var.domain}" : null
+  nat_ip_cidr    = var.whitelist_cidr == ["0.0.0.0/0"] ? [] : formatlist("%s/32", module.vpc.nat_public_ips)
 }
--- a/modules/common/locals.tf
+++ b/modules/common/locals.tf
@@ -2,5 +2,4 @@ locals {
   vpc_name       = format("atlas-%s-vpc", var.environment_name)
   cluster_name   = format("atlas-%s-cluster", var.environment_name)
   ingress_domain = var.domain != null ? "${var.environment_name}.${var.domain}" : null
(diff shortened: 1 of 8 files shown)


modules/common/main.tf
```diff
@@ -33,7 +33,7 @@ module "ingress" {
   # inputs
   ingress_domain              = local.ingress_domain
   enable_ssh_tcp              = var.enable_ssh_tcp
-  load_balancer_access_ranges = var.whitelist_cidr
+  load_balancer_access_ranges = concat(var.whitelist_cidr, local.nat_ip_cidr)
```
nanux, line 36: Would be good to add a comment explaining why we need the concat function here.

modules/common/locals.tf
```diff
@@ -2,4 +2,5 @@ locals {
   vpc_name       = format("atlas-%s-vpc", var.environment_name)
   cluster_name   = format("atlas-%s-cluster", var.environment_name)
   ingress_domain = var.domain != null ? "${var.environment_name}.${var.domain}" : null
+  nat_ip_cidr    = var.whitelist_cidr == ["0.0.0.0/0"] ? [] : formatlist("%s/32", module.vpc.nat_public_ips)
```
nanux, line 5: Can we add unit test for this?

nanux (approved):

modules/common/main.tf
bianchi2: Added a comment

modules/common/locals.tf
bianchi2: Since this value goes into helm_release module, namely into `values` attribute, sadly, it is not available in planned variables for ingress module, so even if I pass a dummy `vpc.nat_public_ips` (i.e. not get it from the module, but pass the module which makes it possible to pass custom values in tests) I can't verify it anywhere.

image

An alternative is to use `set` in helm release (instead of values) which I explored and updated this PR accordingly. Had to move all the logic to helm module. Not sure it's worth one assertion but it's asserted now :)

PR #320, opened 2023-04-27
Add monitoring stack

bianchi2:
This PR makes it possible to optionally deploy https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack with prometheus, node exporter and grafana. Optionally, one can expose Grafana svc as LoadBalancer or use kubectl port-forward to access Grafana UI and build dashboards.

This PR also changes some test data. Currently, unit tests fail in main too because of that. So, names of test instance types have been shortened. Also, since monitoring stack is deployed in common module, eks output is flagged as being sensitive, so adding sensitive attribute to it (to make tests happy).

So far, no e2e tests but we'll have some when jmx metrics is released in Helm charts and it'll be possible to create serviceMonitors and, say, check some request metrics.

## Checklist


test/unittest/test_variables.go
```diff
@@ -46,7 +46,7 @@ var EksWithValidValues = map[string]interface{}{
 	"region":                          "us-east-1",
 	"tags":                            "{}",
 
-	"instance_types":       []string{"instance_type1", "instance_type2"},
+	"instance_types":       []string{"a", "b"},
```
bianchi2, line 49: See: https://github.com/atlassian-labs/data-center-terraform/actions/runs/4825033596/jobs/8595491320#step:7:4478

modules/common/monitoring.tf
```diff
@@ -0,0 +1,55 @@
...
+          loadBalancerSourceRanges = var.whitelist_cidr
+        },
+        persistence = {
+          enabled = true
+          size    = "10Gi"
+        },
+        sidecar = {
+          dashboards = {
+            enabled    = true,
+            label      = "grafana_dashboard",
```
bianchi2, line 32: Grafana sidecar will import dashboards defined in configmaps with this label (in any namespace)

docs/docs/userguide/configuration/CONFIGURATION.md
```diff
@@ -235,6 +235,38 @@ logging_bucket = <LOGGING_S3_BUCKET_NAME>  # default is null
...
+When enabled, Terraform will deploy [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack){.external} Helm chart
+with Prometheus, AlertManager, Node Exporter and Grafana.
+
+By default, Grafana service isn't exposed, and you can login to Grafana at `http://localhost:3000` after running:
+
+```
+kubectl port-forward $grafana-pod 3000:3000 -n kube-monitoring
+```
+
+If you want to expose Grafana service, set `monitoring_grafana_expose_lb` to `true`:
```
jjeongatl, line 256:
```suggestion
If you want to expose Grafana service as `LoadBalancer`, set `monitoring_grafana_expose_lb` to `true`:
```

jjeongatl: I know there is only one way to expose but good to clarify what service type we are using

modules/common/monitoring.tf
```diff
@@ -0,0 +1,55 @@
...
+      },
+      prometheus = {
+        prometheusSpec = {
+          storageSpec = {
+            volumeClaimTemplate = {
+              spec = {
+                accessModes = ["ReadWriteOnce"],
+                resources = {
+                  requests = {
+                    storage = "10Gi"
```
jjeongatl, line 45: 10Gi should be enough for short running instances I assume? Curious to know the growth rate of data with current configuration.

variables.tf
```diff
@@ -134,6 +134,17 @@ variable "enable_https_ingress" {
   default     = true
 }
 
+
+variable "monitoring_enabled" {
+  type    = bool
```
jjeongatl, line 139:
```suggestion
  description = "Enable kube-prometheus-stack for product node monitoring."
  type    = bool
  
```

variables.tf
```diff
@@ -134,6 +134,17 @@ variable "enable_https_ingress" {
...
+  default = false
+}
+
+variable "monitoring_grafana_expose_lb" {
+  type    = bool
```
jjeongatl, line 144:
```suggestion
  description = "Expose Loadbalancer for Grafana."
  type    = bool
```

jjeongatl (approved):

modules/common/monitoring.tf
bianchi2: Good point. Yes, potentially more storage can be required, though it depends on the number of metrics, as well as metrics cardinality and Prometheus retention period.

I have added 2 more variables:

```
prometheus_pvc_disk_size
grafana_pvc_disk_size
```

as well as an info box/comments saying that you can't resize the volume once created, so if you believe your cluster will be busy, override the pvc storage claim when deploying monitoring stack for the first time.

PR #321, opened 2023-05-01
CLIP-1746: Add custom values file overrides

bianchi2:
Currently, if any new variable needs to be passed to Helm release, it should be added to root level variables, then variables in the product module, and finally to values in helm.tf. With changes in this PR, it will be possible to pass a path to a custom values file instead (which is a good alternative to using a local helm chart). Values from config.tfvars will take precedence over those defined in a custom values.yaml.

## Checklist


jjeongatl (approved):
LGTM

PR #325, opened 2023-05-19
Change docker base image

dmika1:
## Pull request description

Decreased ubuntu version to 18.04 inside Dockerfile to pass through apt-get gpg errors

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@
 # AWS_SECRET_ACCESS_KEY="asd123asd123"
 # AWS_ACCESS_KEY_ID="123dsa321asd"
 
-FROM ubuntu:22.04
+FROM ubuntu:18.04
 
 RUN apt-get update \
     && apt-get install -y gnupg software-properties-common curl unzip \


Dockerfile
```diff
@@ -3,19 +3,19 @@
...
 # -v "$PWD/logs:/data-center-terraform/logs" \
-# -it localtf ./scripts/collect_k8s_logs.sh atlas-cluster-name-cluster us-east-2 k8s_logs
+# -it atlassianlabs/terraform:main ./scripts/collect_k8s_logs.sh atlas-cluster-name-cluster us-east-2 k8s_logs
 
 # In those example aws_envs should contain AWS variables needed for authorization like:
 # AWS_SECRET_ACCESS_KEY="asd123asd123"
 # AWS_ACCESS_KEY_ID="123dsa321asd"
 
-FROM ubuntu:22.04
+FROM ubuntu:18.04
```
ometelytsia, line 18: @bianchi2 we need some help here. We could not build an image on bamboo CI with ubuntu 22.04 due to gpg error.
@dmika1 could you please add more details?

nanux: Do we have a log file?

dmika1: The problem is already solved. No ubuntu downgrade needed and this PR can be rejected. I'll create today PR with just modified docker image name from localtf to atlassianlabs/terraform.

PR #328, opened 2023-06-05
Use name prefix for launch template. Delete IAM roles and policies

bianchi2:
Changes in this PR:

* use `name_prefix` for launch template instead of a name (always good to use it to avoid already exist errors in cases when uninstall failed and the infra was cleaned up with some script)
* add delete iam roles and policies functions to AWS cleanup script
* add delete launch templates function to AWS cleanup script

## Checklist

### DIFF
--- a/aws_clean.py
+++ b/aws_clean.py
@@ -376,6 +376,47 @@ def delete_hosted_zones(service_name):
             client.delete_hosted_zone(Id=hz['Id'])
 
 
+def delete_iam_roles(service_name):
+    logging.info('Deleting IAM roles')
+    client = boto3.client('iam')
+    response = client.list_roles(MaxItems=1000)
+    matching_roles = [role['RoleName'] for role in response['Roles'] if
+                      role['RoleName'].startswith('atlas-' + service_name)]
+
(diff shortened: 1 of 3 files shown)


aws_clean.py
```diff
@@ -376,6 +376,65 @@ def delete_hosted_zones(service_name):
             client.delete_hosted_zone(Id=hz['Id'])
 
 
+def delete_iam_roles(service_name):
+    logging.info('Deleting IAM roles')
+    client = boto3.client('iam')
+    response = client.list_roles(MaxItems=1000)
```
errcode1202, line 382: Is the assumption here that there should never be more than `1000` roles at any given time? Reason I ask is because it looks like if there are say, `1001` roles, only `1000` will be nuked. In saying that, since you manually cleaned these up recently I'd like to think that now this script is in place we should never be in a situation again where we have > `1000` to clean up.

errcode1202 (approved):
❤️

aws_clean.py
bianchi2: You're right. Let me introduce some pagination instead.

PR #341, opened 2023-10-23
Get snapshots information from json file

bianchi2:
This PR keeps an existing way to read snapshot IDs from tfvars file but adds an alternative method - parse json file and get EBS and RDS snapshots IDs from it. Depending on the product, version and dataset size snapshots IDs are passed to each product module.

Snapshots json takes precedence over existing snap variables if they are defined.

## Checklist

### DIFF
--- a/docs/docs/userguide/configuration/CONFIGURATION.md
+++ b/docs/docs/userguide/configuration/CONFIGURATION.md
@@ -281,6 +281,41 @@ grafana_pvc_disk_size = "20Gi"
     to override the default Prometheus 10Gi PVC storage request when creating enabling monitoring for the first time.
     [AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html){.external}.
 
+### Snapshot Configuration
+
+It is possible to restore DC products from snapshots. Each DC product requires a valid public RDS and EBS (shared-home) snapshot defined by the following variables:
+
+```
+<product>_db_snapshot_id
+```
+
+```
+<product>_shared_home_snapshot_id
+```
+
+Note that COnfluence and Crowd also require a build number:
+
+```
+confluence_db_snapshot_build_number = "8017"
+crowd_db_snapshot_build_number = "5023"
+```
+
+Snapshots must be public and exist in the target region.
+
+!!! info "Snapshots JSON File"
+    
+    It is also possible to use a special snapshots JSON file with pre-defined snapshot ID and build numbers for all products for both small and large dataset sizes.
+
+    You can find example JSON in `test/dcapt-snapshots.json`. To use snapshots JSON rather than dedicated environment variables, set in `config.tfvars`:
+
+    ```
+    snapshots_json_file_path = "test/dcapt-snapshots.json"
+    ```
+    
+    If `snapshots_json_file_path` snapshot variables defined in `config.tfvars` are ignored.
+    
+    Only use snapshots JSON suggested by [DCAPT team](https://github.com/atlassian/dc-app-performance-toolkit){.external}. 
+
 ## Product specific configuration
 
 === "Bamboo"


config.tfvars
```diff
@@ -85,6 +85,9 @@ max_cluster_capacity = 5
 # List of addtional namespaces to be created in the cluster
 #additional_namespaces = ["extra_namespace"]
 
+# Path to a JSON file with EBS and RDS snapshot IDs
+#snapshots_json_file_path = "test/dcapt-snapshots.json"
```
nanux, line 89: /nit we should be consistent with the space after the `#` comment symbol. I think leaving a space in is better for readability and removing the space is more convenient as I can just uncomment the line and don't delete the leading space.

I am leaning towards leaving a space

config.tfvars
```diff
@@ -215,6 +218,9 @@ jira_db_iops                 = 1000
 # Set `null` if the snapshot does not have a default db name.
 jira_db_name = "jira"
 
+# Dataset size. Used only when snapshots_json_file_path is defined. Defaults to large
+# jira_dataset_size = "small"
```
nanux, line 222:
```suggestion
# jira_dataset_size = "large"
```

When we default to something, we should keep the value here.

config.tfvars
```diff
@@ -310,6 +316,9 @@ confluence_db_iops                 = 1000
 # Set `null` if the snapshot does not have a default db name.
 confluence_db_name = "confluence"
 
+# Dataset size. Used only when snapshots_json_file_path is defined. Defaults to large
+# confluence_dataset_size = "small"
```
nanux, line 320:
```suggestion
# confluence_dataset_size = "large"
```

config.tfvars
```diff
@@ -420,6 +429,9 @@ bitbucket_db_name = "bitbucket"
 #bitbucket_elasticsearch_storage         = "<REQUESTS_STORAGE>"
 #bitbucket_elasticsearch_replicas        = "<NUMBER_OF_NODES>"
 
+# Dataset size. Used only when snapshots_json_file_path is defined. Defaults to large
+# bitbucket_dataset_size = "small"
```
nanux, line 433:
```suggestion
# bitbucket_dataset_size = "large"
```

config.tfvars
```diff
@@ -590,6 +602,9 @@ crowd_db_name                 = "crowd"
 # if you encounter such an issue. This will apply to Crowd pods.
 #crowd_termination_grace_period = 0
 
+# Dataset size. Used only when snapshots_json_file_path is defined. Defaults to large
+# crowd_dataset_size = "small"
```
nanux, line 606:
```suggestion
# crowd_dataset_size = "large"
```

locals.tf
```diff
@@ -17,4 +17,100 @@ locals {
   local_bamboo_chart_path     = var.local_helm_charts_path != "" && var.bamboo_install_local_chart ? "${var.local_helm_charts_path}/bamboo" : ""
   local_agent_chart_path      = var.local_helm_charts_path != "" && var.bamboo_install_local_chart ? "${var.local_helm_charts_path}/bamboo-agent" : ""
   local_crowd_chart_path      = var.local_helm_charts_path != "" && var.crowd_install_local_chart ? "${var.local_helm_charts_path}/crowd" : ""
+
+  #  snapshots_json = var.snapshots_json_file_path != "" ? jsondecode(file(var.snapshots_json_file_path)) : jsondecode("{\"jira\": {\"versions\":[]},\"confluence\": {\"versions\":[]},\"crowd\": {\"versions\":[]},\"bitbucket\": {\"versions\":[]}}")
```
nanux, line 21: Do we need this?

outputs.tf
```diff
@@ -111,3 +111,43 @@ output "confluence_s3_bucket" {
...
+
+output "confluence_rds_snapshot" {
+  value = local.confluence_rds_snapshot_id
+}
+
+output "confluence_ebs_snapshot" {
+  value = local.confluence_ebs_snapshot_id
+}
+
+output "confluence_db_snapshot_build_number" {
```
nanux, line 131: Only Confluence and Crowd require build number?

install.sh
```diff
@@ -147,6 +151,9 @@ pre_flight_checks() {
       fi
     fi
     RDS_SNAPSHOT_ID=$(get_variable ${RDS_SNAPSHOT_VAR} "${CONFIG_ABS_PATH}")
+    if [ "${SNAPSHOTS_JSON_FILE_PATH}" ]; then
+      RDS_SNAPSHOT_ID=$(cat ${SNAPSHOTS_JSON_FILE_PATH} | jq ".${PRODUCT}.versions[] | select(.version == \"${PRODUCT_VERSION}\") | .data[] | select(.size == \"large\" and .type == \"rds\") | .snapshots[] | .[\"${REGION}\"]" | sed 's/"//g')
```
nanux, line 155: Doesn't this always read the `large` dataset from the JSON file and uses that?

variables.tf
```diff
@@ -394,6 +404,16 @@ variable "jira_install_local_chart" {
   type        = bool
 }
 
+variable "jira_dataset_size" {
+  description = "The size of the dataset to restore from"
+  type        = string
+  default     = "large"
+  validation {
+    condition     = var.jira_dataset_size == null || can(regex("large|small", var.jira_dataset_size))
```
nanux, line 412:
```suggestion
    condition     = var.jira_dataset_size == null || contains(["large", "small"], var.jira_dataset_size)
```

I find this easier to read.

test/dcapt-snapshots.json
nanux, line 1: Is this the snapshot file that is going to be used with DCAPT or we are distributing the DCAPT snapshot json via DCAPT framework?

locals.tf
```diff
@@ -17,4 +17,100 @@ locals {
...
+  snapshots_json = var.snapshots_json_file_path != "" ? jsondecode(file(var.snapshots_json_file_path)) : null
```
nanux, line 22: Can we put a sentence or two describing the filtering process and how the snapshot values are determined?

I had to read it a couple of times to see how this is used, and I am sure I will forget it before I will see this code again :)

nanux (approved):

PR #344, opened 2023-11-14
Make crowd test more stable

bianchi2:
Just a few small things that will make crowd tests more stable. During the tests, we need to create a new user directory in Btibucket. Since there's no REST API for that, we first need to get atl_token by making a get call and parsing html response. We run 3 instances of Bitbucket, and atl_token is a per tomcat thing.That's why we scale Btibucket to 1 first. Without a short wait, occasionally Bitbucket does not respond with atl_token.  After the scaledown, let's give Kube and Bitbucket some time to get back to a normal state. Also, let's make an additional call if the initial one did not return atl_token (network glitch etc).

[Tested 13 runs](https://github.com/atlassian-labs/data-center-terraform/actions/runs/6795657549) :) Not a single failure.

Also, increasing tests timeout as sometimes due to networking conditions and how fast AWS is on a given day, the tests are timing out.

## Checklist


yzha645 (approved):

nanux (approved):
:nice:

PR #345, opened 2023-11-19
Create an optional deployment with a privileged dind container

bianchi2:
This PR adds an optional deployment with docker-in-docker privileged container that can be used to run Jmeter and Selenium tests.

## Checklist

### DIFF
--- a/docs/docs/userguide/configuration/DCAPT_CONFIGURATION.md
+++ b/docs/docs/userguide/configuration/DCAPT_CONFIGURATION.md
@@ -1,0 +1,43 @@
+# DCAPT Configuration
+
+Besides deploying Atlassian DC products, it is possible to spin up a testing environment to run [Data Center App Performance Toolkit tests](https://github.com/atlassian/dc-app-performance-toolkit/tree/master).
+
+## Create DCAPT Test Deployment
+
+Create an additional deployment for DCAPT Jmeter and Selenium test environment:
+
+```shell
+start_test_deployment = true
+```
+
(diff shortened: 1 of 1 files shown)


config.tfvars
```diff
@@ -141,6 +141,31 @@ max_cluster_capacity = 5
 # are passed to Helm chart. Variables from config.tfvars take precedence over those defined in a custom values.yaml.
 # monitoring_custom_values_file = "/path/to/values.yaml"
 
+################################################################################
+# DCAPT JMeter and Selenium Deployment Settings
+################################################################################
+
+# Create deployment with a docker-in-docker privileged container. Defaults to false
+# start_test_deployment = true
```
nanux, line 149: I thought we were setting the commented values to the defaults.

variables.tf
```diff
@@ -1317,3 +1317,47 @@ variable "confluence_s3_attachments_storage" {
   type        = bool
   default     = false
 }
+
+# test deployment configuration
+
+variable "start_test_deployment" {
+  description = "Whether to start DCAPT Jmeter and Selenium deployment"
```
nanux, line 1324: Should we say something like "Deploy necessary resources to start DCAPT testing"? So it is potentially flexible in future.

nanux (approved):

nanux:
Should we create DCAPT section under https://atlassian-labs.github.io/data-center-terraform/userguide/configuration/CONFIGURATION/ so we can start documenting DCAPT specifics there?

PR #349, opened 2023-12-14
Check product license

bianchi2:
This PR adds a simple license check. There are 2 checks:

* check if product_license is an empty string, undefined or a multi line string
* check if product_license is a placeholder (i.e. it has never been updated)

In bash it is tricky to extract a multi line value for a variable, but it's good enough to know that something is wrong with the license. Below is example output. 

empty, undefined or multi-line:

```
2023-12-14 14:40:39 [ERROR] License is undefined or malformed. Please check 'bitbucket_license' value in '/Users/users/Downloads/dcapt.tfvars'
2023-12-14 14:40:39 [ERROR] It is possible that 'bitbucket_license' is defined but it is a multi line string
2023-12-14 14:40:39 [ERROR] If that's the case remove all new lines to convert it to a one-liner
```

placeholder:

```
2023-12-14 14:34:20 [ERROR] License placeholder is unchanged. Please check 'bitbucket_license' value in '/Users/user/Downloads/dcapt.tfvars'
2023-12-14 14:34:20 [ERROR] Current license key: 'bitbucket-license'
2023-12-14 14:34:20 [ERROR] Generate a new license at https://my.atlassian.com/
```

`-l` disables license check which is ON by default. This flag is added to e2e tests since license vars are defined with TF_VAR rather than in a config file.

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -46,16 +46,16 @@ EOF
   HELP_FLAG=
   FORCE_FLAG=
   CLEAN_UP_FLAG=
-  PRE_FLIGHT_FLAG=
-  LICENSE_TEST_FLAG=
+  SKIP_PRE_FLIGHT_FLAG=
+  SKIP_LICENSE_TEST_FLAG=
   while getopts hfdpl?c: name ; do
       case $name in
       h)    HELP_FLAG=1; show_help;;  # Help
       c)    CONFIG_FILE="${OPTARG}";; # Config file name to install - this overrides the default, 'config.tfvars'
       f)    FORCE_FLAG="-f";;         # Auto-approve
       d)    CLEAN_UP_FLAG="-d";;      # Run cleanup script before install
-      p)    PRE_FLIGHT_FLAG="-p";;    # Skip pre-flight checks to test compatibility of EBS and RDS snapshots if any
-      l)    LICENSE_TEST_FLAG="-l";;  # Skip license checks
+      p)    SKIP_PRE_FLIGHT_FLAG="-p";;    # Skip pre-flight checks to test compatibility of EBS and RDS snapshots if any
+      l)    SKIP_LICENSE_TEST_FLAG="-l";;  # Skip license checks
       ?)    log "Invalid arguments." "ERROR" ; show_help
       esac
   done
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -45,13 +47,15 @@ EOF
...
-  while getopts hfdp?c: name ; do
+  LICENSE_TEST_FLAG=
+  while getopts hfdpl?c: name ; do
       case $name in
       h)    HELP_FLAG=1; show_help;;  # Help
       c)    CONFIG_FILE="${OPTARG}";; # Config file name to install - this overrides the default, 'config.tfvars'
       f)    FORCE_FLAG="-f";;         # Auto-approve
       d)    CLEAN_UP_FLAG="-d";;      # Run cleanup script before install
       p)    PRE_FLIGHT_FLAG="-p";;    # Skip pre-flight checks to test compatibility of EBS and RDS snapshots if any
+      l)    LICENSE_TEST_FLAG="-l";;  # Skip license checks
```
nanux, line 58:
```suggestion
      l)    SKIP_LICENSE_TEST_FLAG="-l";;  # Skip license checks
```
I would expect `LICENSE_TEST_FLAG` to actually enable the test.

nanux (approved):

install.sh
```diff
@@ -106,6 +110,26 @@ pre_flight_checks() {
...
+      log "Checking ${PRODUCT} license"
+      LICENSE_ENV_VAR=${PRODUCT}'_license'
+      LICENSE_TEXT=$(get_variable ${LICENSE_ENV_VAR} "${CONFIG_ABS_PATH}")
+      if [ -z "$LICENSE_TEXT" ]; then
+        log "License is undefined or malformed. Please check '${LICENSE_ENV_VAR}' value in '${CONFIG_ABS_PATH}'" "ERROR"
+        log "It is possible that '${LICENSE_ENV_VAR}' is defined but it is a multi line string" "ERROR"
+        log "If that's the case remove all new lines to convert it to a one-liner" "ERROR"
+        exit 1
+      fi
+      if [ "${LICENSE_TEXT}" == ${PRODUCT}-license ]; then
```
yzha645, line 124: The placeholder value `${PRODUCT}-license` comes from [config file in DCAPT repo](https://github.com/atlassian/dc-app-performance-toolkit/blob/master/app/util/k8s/dcapt-small.tfvars#L78C23-L78C23). 
If they change the placeholder value, which is unlikely, we'll need to change here as well.

yzha645 (approved):

PR #351, opened 2023-12-18
Optimize Terraform plan to decrease overall deployment time

bianchi2:
# What This PR Is All About

This PR moves RDS module from products to root level. Currently, product modules are called from `dc-instrastructure.tf`, and each module [calls](https://github.com/atlassian-labs/data-center-terraform/blob/main/modules/products/jira/main.tf#L32) a `database` module. This causes 2 issues:

* RDS becomes dependent on EKS and thus it is created after EKS with deps is created and before Helm charts are deployed
* this makes terraform plan overly complicated and it takes ~2.5-3 min to process it

By dynamically creating RDS for each product in `products` list we achieve a significantly faster deployment because:

*  RDS is ready by the time EKS is created and configured and Helm charts are ready to be deployed. On average RDS creation takes roughly 5 minutes and up to 7-8 minutes if it's restored from a snapshot
* terraform plan becomes a lot lighter and there are fewer dependencies

RDS module is passed to each product module to get endpoint and master credentials. RDS module itself uses count and dynamic variable generation. To achieve it a few new variables has been introduced to locals.

# How Much Faster Is It?

A lot. In addition to up to a 8 minute gain due to creating RDS at the same time with EKS, terraform plan becomes much lighter.

On main branch now (product = confluence, with rds and ebs snapshots):

```
$ :~/projects/terraform/data-center-terraform (main) { time terraform plan -var-file=/Users/user/dcapt.tfvars; } 2>&1 | grep real
real	2m42.043s
```
And on PR branch:

```
$ :~/projects/terraform/data-center-terraform (create-database-with-eks) { time terraform plan -var-file=/Users/user/dcapt.tfvars; } 2>&1 | grep real
real	0m23.407s
```
So, almost 2.5 min. Add it to 7-8 min upfront RDS creation gain, and we have solid **10 minutes**! When running terraform graph to save plan as svg, the resulting files differ in size (which explains why terraform plan is so much faster):

image

e2e tests now take <50 min while they typically took well over an hour
See: https://github.com/atlassian-labs/data-center-terraform/actions/runs/7253992209/job/19761764216

See inline comments with more details on changes.

## Checklist

### DIFF
--- a/dc-infrastructure.tf
+++ b/dc-infrastructure.tf
@@ -124,14 +124,14 @@ module "jira" {
   count      = local.install_jira ? 1 : 0
   depends_on = [module.base-infrastructure]
 
-  environment_name        = var.environment_name
-  namespace               = module.base-infrastructure.namespace
-  vpc                     = module.base-infrastructure.vpc
-  eks                     = module.base-infrastructure.eks
-  rds                     = module.database[index(var.products, "jira")]
-  ingress                 = module.base-infrastructure.ingress
+  environment_name = var.environment_name
+  namespace        = module.base-infrastructure.namespace
+  vpc              = module.base-infrastructure.vpc
+  eks              = module.base-infrastructure.eks
+  rds              = module.database[index(var.products, "jira")]
+  ingress          = module.base-infrastructure.ingress
 
-  db_snapshot_id          = local.jira_rds_snapshot_id
+  db_snapshot_id = local.jira_rds_snapshot_id
 
   replica_count            = var.jira_replica_count
   installation_timeout     = var.jira_installation_timeout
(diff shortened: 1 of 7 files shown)


dc-infrastructure.tf
```diff
@@ -118,6 +136,7 @@ module "jira" {
   namespace               = module.base-infrastructure.namespace
   vpc                     = module.base-infrastructure.vpc
   eks                     = module.base-infrastructure.eks
+  rds                     = module.database[index(var.products, "jira")]
   ingress                 = module.base-infrastructure.ingress
   db_major_engine_version = var.jira_db_major_engine_version
```
yzha645, line 141: Since rds module is no longer in product module, we don't really need `db_major_engine_version` here anymore.

yzha645 (approved):

PR #355, opened 2023-12-25
Support for optional local-home EBS snapshots

bianchi2:
# Problem Statement

When Jira and Confluence are restored from large datasets, it takes 10+ minutes to recover cache/indexes because local home PVC is created from a template, and thus it is created empty. Both Jira and Confluence will look at local home and if indexes aren't found there, they would recover them from shared home.

By pre-warming local-home PVs, the overall deployment time is 2x faster, because no index recovery is required.

# Solution
This PR adds optional EBS volume (restored from a snapshot), a PV and a PVC (that are mutually bound) to Jira and Confluence modules. Helm release becomes dependent on the local home PVC. The number of local home PVCs == the number of desired replicas.

## Checklist

### DIFF
--- a/modules/products/confluence/local_home_storage.tf
+++ b/modules/products/confluence/local_home_storage.tf
@@ -1,8 +1,14 @@
+data "aws_ebs_snapshot" "local_home_snapshot" {
+  count = var.local_home_snapshot_id != null ? var.replica_count : 0
+  snapshot_ids           = [var.local_home_snapshot_id]
+  most_recent            = true
+}
+
 resource "aws_ebs_volume" "local_home" {
   count             = var.local_home_snapshot_id != null ? var.replica_count : 0
   availability_zone = var.eks.availability_zone
   snapshot_id       = var.local_home_snapshot_id
-  size              = tonumber(regex("\\d+", var.local_home_size))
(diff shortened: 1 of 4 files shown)


modules/products/confluence/local_home_storage.tf
```diff
@@ -0,0 +1,51 @@
+resource "aws_ebs_volume" "local_home" {
+  count             = var.local_home_snapshot_id != null ? var.replica_count : 0
+  availability_zone = var.eks.availability_zone
+  snapshot_id       = var.local_home_snapshot_id
+  size              = tonumber(regex("\\d+", var.local_home_size))
```
nanux, line 5: Would be possible to get the size from the snapshot instead? I.e. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_snapshot#volume_size

What will happen if vendor set the `local_home_size` smaller to the snapshot size?

nanux (approved):
Good improvement!

modules/products/confluence/local_home_storage.tf
nanux: Nice update. LGTM.

PR #356, opened 2024-01-02
Deploy NFS server immediately after EKS module is ready

bianchi2:
This is another PR in a series of PRs aimed at reducing infrastructure creation time.

Currently, NFS module is called from a product module ([example](https://github.com/atlassian-labs/data-center-terraform/blob/main/modules/products/confluence/main.tf#L16)). It means that Terraform waits until common module is **fully** created. However, there's no need to wait, say, for ingress deployment (which takes ~3 min) and deployment of any other tools like cluster autoscaler.

This PR takes NFS module out of product modules and creates NFS module dynamically for each product. This flattening results in:
* at least 1,5 min faster deployment time (the amount of time Terraform needs to create EBS, PV, PVC and deploy nfs-server Helm chart)
* flattened terraform plan - Terraform need less time to create one (at least 2-3 seconds faster). Terraform graph has also become lighter.

```
user@user:~/projects/data-center-terraform (main) for _ in {1..5}; do \
{ time terraform plan -var-file=dcapt.tfvars \
-refresh=false -refresh=false -lock=false; } 2>&1 | grep real; done
real	0m19.462s
real	0m18.557s
real	0m18.634s
real	0m18.352s
real	0m17.921s
user@user:~/projects/data-center-terraform (main) git co nfs-server-deployment
Switched to branch 'nfs-server-deployment'
user@user:~/projects/data-center-terraform (nfs-server-deployment) for _ in {1..5}; do \
{ time terraform plan -var-file=dcapt.tfvars \
-refresh=false -refresh=false -lock=false; } 2>&1 | grep real; done
real	0m16.396s
real	0m16.313s
real	0m16.968s
real	0m16.552s
real	0m16.646s
```

## Checklist

### DIFF
--- a/test/e2etest/confluence_test.go
+++ b/test/e2etest/confluence_test.go
@@ -40,7 +40,7 @@ func assertS3BucketPermissions(t *testing.T, testConfig TestConfig) {
 
 	// install aws cli
 	println("Installing AWS CLI ...")
-	command := []string{"apt-get update; apt-get install unzip -y; curl \"https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip\" -o \"awscliv2.zip\"; unzip awscliv2.zip; ./aws/install"}
+	command := []string{"apt-get update; apt-get install unzip -y; curl \"https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip\" -o \"awscliv2.zip\"; unzip -qq awscliv2.zip; ./aws/install"}
 	_, kubectlError := k8s.RunKubectlAndGetOutputE(t, kubectlOptions, append(execArgs, command...)...)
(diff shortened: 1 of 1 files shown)


yzha645 (approved):

nanux (approved):
LGTM.

I wonder if we made NFS part of product just for simplicity reasons.

Is having a single NFS with different mounts bad idea? I can see this as a cleaner solution, I just wonder if it would speed it up even more.

nanux:
> @nanux I think we can get away with just one nfs server but it will make the restore process more complicated. Also, since nfs server per product is being created when other EKS tooling is being deployed, all nfs server are ready by the time EKS is ready for Helm charts. So, no time wins here.

That does make sense, thanks for explanation.

PR #358, opened 2024-01-04
Force delete cluster if terraform destroy failed

bianchi2:
This PR will automatically run dcapt cleanup script if uninstall.sh fails (terraform destroy in particular).

Also, @ometelytsia has suggested that we make S3 bucket and DynamoDB names unique, thus adding environment_name. S3 bucket name can't exceed 63 characters, so cutting it if it does.

https://github.com/atlassian-labs/data-center-terraform/actions/runs/7405211391/job/20147833366#step:13:12959 bucket and DynamoDB deletion.

## Checklist

### DIFF
--- a/scripts/generate-variables.sh
+++ b/scripts/generate-variables.sh
@@ -60,7 +60,6 @@ set_variables() {
 
   # Generates the unique dynamodb table names for the deployment lock ( convert all '-' to '_' )
   DYNAMODB_TABLE="atl_dc_${ENVIRONMENT_NAME//-/_}_${REGION//-/_}_${AWS_ACCOUNT_ID}_tf_lock"
-  DYNAMODB_TABLE=${DYNAMODB_TABLE:0:MAX_LENGTH}
 
   BACKEND_TF="${ROOT_PATH}/terraform-backend.tf"
   TFSTATE_LOCALS="${ROOT_PATH}/modules/tfstate/tfstate-locals.tf"


scripts/generate-variables.sh
```diff
@@ -53,11 +53,14 @@ set_variables() {
...
-  S3_BUCKET="atlassian-data-center-${REGION}-${AWS_ACCOUNT_ID}-tf-state"
+  MAX_LENGTH=63
+  S3_BUCKET="atl-dc-${ENVIRONMENT_NAME}-${REGION}-${AWS_ACCOUNT_ID}-tfstate"
+  S3_BUCKET=${S3_BUCKET:0:MAX_LENGTH}
   BUCKET_KEY="${ENVIRONMENT_NAME}"
 
   # Generates the unique dynamodb table names for the deployment lock ( convert all '-' to '_' )
-  DYNAMODB_TABLE="atlassian_data_center_${REGION//-/_}_${AWS_ACCOUNT_ID}_tf_lock"
+  DYNAMODB_TABLE="atl_dc_${ENVIRONMENT_NAME//-/_}_${REGION//-/_}_${AWS_ACCOUNT_ID}_tf_lock"
+  DYNAMODB_TABLE=${DYNAMODB_TABLE:0:MAX_LENGTH}
```
yzha645, line 63: Do we need to restrain dynamodb table name length to 63 as well? The max is 255. 
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateTable.html#API_CreateTable_RequestParameters

yzha645 (approved):

PR #365, opened 2024-02-12
Update EKS to 1.29

bianchi2:
1.29 is the latest supported EKS version. End of standard support: March 23, 2025

## Checklist


yzha645:
Another comment to be updated: https://github.com/atlassian-labs/data-center-terraform/blob/main/config.tfvars#L16

yzha645 (approved):

PR #369, opened 2024-02-20
Fix index for local snapshot data

bianchi2:
This PR fixes 2 things:

1) It's safe to always grab the first instance of data since we always have at least 1 replica.
2) Set `ATL_FORCE_CFG_UPDATE` to true to overwrite Jira's dbconfig.xml if local_home_snapshot is defined. This will force image entrypoint to generate the file from scratch. Same for Confluence cfg.xml


nanux (approved):

nanux:
What was the behaviour without the fixes?

PR #375, opened 2024-03-25
Add optional crowdstrike deployment to instance user data

bianchi2:
This PR adds an optional crowdstrike installation to a user data script (that is executed after the ec2 instance starts). 

Defining `crowdstrike_secret_name` enables crowdstrike installation. There are other obligatory variables:

```
# name of the AWS secret holding cid and token
# crowdstrike_secret_name = "crowdstrike-secret"

# name of the kms key which is used to decrypt the secret
# crowdstrike_kms_key_name = "kms-key"

# AWS account that shares crowdstrike resources
# crowdstrike_aws_account_id = "1234567890"
```
How the script works:
* falcon sensor binaries is fetches from a pre-defined s3 bucket
* cid and token are fetched from SecretsManager
* crowdstrike is configured and started

## Checklist

### DIFF
--- a/config.tfvars
+++ b/config.tfvars
@@ -89,6 +89,22 @@ max_cluster_capacity = 5
 # snapshots_json_file_path = "test/dcapt-snapshots.json"
 
 ################################################################################
+# Crowdstrike settings. Atlassian only!
+################################################################################
+
+# name of the AWS secret holding cid and token
+# crowdstrike_secret_name = "crowdstrike-secret"
+
+# name of the kms key which is used to decrypt the secret
+# crowdstrike_kms_key_name = "kms-key"
+
+# AWS account that shares crowdstrike resources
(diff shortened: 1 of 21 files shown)


config.tfvars
```diff
@@ -88,6 +88,22 @@ max_cluster_capacity = 5
...
+# crowdstrike_secret_name = "crowdstrike-secret"
+
+# name of the kms key which is used to decrypt the secret
+# crowdstrike_kms_key_name = "kms-key"
+
+# AWS account that shares crowdstrike resources
+# crowdstrike_aws_account_id = "1234567890"
+
+# falcon sensor version  version
+# falcon_sensor_version = "7.10.0-16303"
```
yzha645, line 105: Do we need to update the version every now and then?

yzha645 (approved):

PR #411, opened 2024-05-23
Remove rds 10 db engine which is depricated

ometelytsia:
## Pull request description

## Checklist


ometelytsia:
> We should probably get rid of Postgres 11 as well. It's deprecated/not supported in most DC products I think.

@bianchi2 
Jira 9.4 LTS supports Postgres 11. But I'm not sure if we want to keep this support. E.g. app partners usually stick to our defaults, so db would be selected by snapshot used.

One more question, maybe we need to add 15 and 16 versions? Confluence team is already working on Confluence 9 support for PG 16.

PR #416, opened 2024-07-22
CLIP-1883: Role permissions

yzha645:
## Pull request description

This PR provided a guided set of policy statements for the AWS IAM role that runs the Terraform deployment. With more clearly defined services and actions in the policy, users no longer need admin permission for the deployment. 

Also, two scripts are added to facilitate testing for future use, i.e. script to assume a role, and script to update policy statement using CLI and policy document instead of doing the update manually via AWS console.  

## Checklist

### DIFF
--- a/permissions/assume-role.sh
+++ b/permissions/assume-role.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+# This script is useful when you want to do role chaining.
+# Make sure you've already authenticated into AWS CLI before running this script.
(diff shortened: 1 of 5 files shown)


yzha645:
> @yzha645 do we have a follow-up ticket to update CI to use this new role? We need an automatic way to detect when this role got broken or obsolete.

Hey Alex, yes, I'll update the role used for CI later in [CLIP-1903](https://bulldog.internal.atlassian.com/browse/CLIP-1903).

PR #423, opened 2024-07-29
Update versions of Terraform providers and modules and bump EKS version

bianchi2:
We have been using the same version of eks module and aws provider for quite some time. Existing versions are > 2 years old. We need to bump versions to keep them up to date to be able to use new features.

There is one breaking change for EKS. Authentication mechanism now uses access entries, so eks_additional_roles may look like this now:

```
eks_additional_roles = {
  user = {
    kubernetes_group = []
    principal_arn    = "arn:aws:iam::121212121212:role/test-policy-role"
    policy_associations = {
      admin = {
        policy_arn = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
        access_scope = {
          namespaces = []
          type       = "cluster"
        }
      }
    }
  }
}
```
This is an example of granting an admin access to a role, however, it is possible to grant read-only or admin role for a selected namespace.

Also, bumping versions of helm charts and other providers like helm and kubernetes, and adding a few missing permissions to the role.

## Checklist

### DIFF
--- a/permissions/policy.json
+++ b/permissions/policy.json
@@ -73,7 +73,6 @@
         "eks:DeleteAccessEntry",
         "eks:DescribeAddonVersions",
         "eks:DescribeAccessEntry",
-        "eks:TagResource",
         "eks:AssociateAccessPolicy",
         "eks:DisassociateAccessPolicy",
         "eks:ListAssociatedAccessPolicies"


modules/AWS/eks/main.tf
```diff
@@ -36,22 +36,37 @@ module "eks" {
...
+  access_entries                           = local.iam_access_entries
+
+  cluster_endpoint_public_access = true
 
   # Enables IAM roles for service accounts - required for autoscaler and potentially Atlassian apps
   # https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
   enable_irsa              = true
   iam_role_use_name_prefix = false
 
+  # we won't use ksm key to entrypt secrets in etcd
```
yzha645, line 63:
```suggestion
  # we won't use kms key to encrypt secrets in etcd
```

permissions/policy.json
```diff
@@ -67,7 +68,15 @@
         "eks:CreateCluster",
         "eks:TagResource",
         "eks:UntagResource",
-        "eks:ListTagsForResource"
+        "eks:ListTagsForResource",
+        "eks:CreateAccessEntry",
+        "eks:DeleteAccessEntry",
+        "eks:DescribeAddonVersions",
+        "eks:DescribeAccessEntry",
+        "eks:TagResource",
```
yzha645, line 76: Duplicated. See line 69.

yzha645 (approved):

PR #425, opened 2024-08-09
Fix missing permissions: dynamodb:ListTables

ometelytsia:
## Pull request description

Terminate scrips has list dynamodb tables operation: 
https://github.com/atlassian/dc-app-performance-toolkit/blob/dev/app/util/k8s/terminate_cluster.py#L956C16-L956C45

Also, this command fails if there is resource restriction.

## Checklist

### DIFF
--- a/permissions/policy1.json
+++ b/permissions/policy1.json
@@ -92,6 +92,7 @@
         "dynamodb:DescribeTable",
         "dynamodb:DescribeContinuousBackups",
         "dynamodb:CreateTable",
+        "dynamodb:ListTables",
         "dynamodb:TagResource",
         "dynamodb:UntagResource",
         "dynamodb:ListTagsOfResource",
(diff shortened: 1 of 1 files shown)


permissions/policy1.json
```diff
@@ -101,7 +102,7 @@
         "dynamodb:UpdateItem"
       ],
       "Resource": [
-        "arn:aws:dynamodb:*:123456789012:table/atl_dc_*_tf_lock"
+        "arn:aws:dynamodb:*:123456789012:table/*"
```
yzha645, line 105: Probably better to have `"dynamodb:ListTables"` in a standalone statement if we're giving `*` resource?

ometelytsia: @yzha645 done

yzha645 (approved):

PR #435, opened 2024-10-15
Manually scale down Confluence/Jira statefulsets

bianchi2:
# Problem Statement

When scaling down Jira/Confluence with pre-created local-home PVs, Terraform destroys PVC/PV/EBS first and only then updates Helm release. It happens because local-home vols depend on Helm charts and need to be created first. PVC is stuck in terminating state if it's being deleted when the pod is still running.

This PR adds a script that will identify if terraform apply is a scale down event and scale down StatefulSet (and wait until pods are gone). Then Terraform will proceed with deleting local home PVC, PV and EBS which happens almost instantly.

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -393,6 +393,9 @@ set_current_context_k8s() {
   fi
 }
 
+# this function will call an external script that will check if existing StatefulSet replicas > desired
+# StatefulSet replicas (as stated in tfvars) and scale down to desired StatefulSet replicas to prevent Terraform
+# from deleting PVCs attached to running pods
 scale_down() {
     set +e
     PRODUCTS=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep "products" | sed 's/ //g')


install.sh
```diff
@@ -393,6 +393,32 @@ set_current_context_k8s() {
...
 
+scale_down() {
+    set +e
+    PRODUCTS=$(grep -o '^[^#]*' "${CONFIG_ABS_PATH}" | grep "products" | sed 's/ //g')
+    PRODUCTS="${PRODUCTS#*=}"
+    PRODUCTS_ARRAY=($(echo $PRODUCTS | sed 's/\[//g' | sed 's/\]//g' | sed 's/,/ /g' | sed 's/"//g'))
+    if echo "$PRODUCTS" | grep -qE 'jira|confluence'; then
+      SNAPSHOTS_JSON_FILE_PATH=$(get_variable 'snapshots_json_file_path' "${CONFIG_ABS_PATH}")
+      if [ "${PATH}" ]; then
+        local EKS_PREFIX="atlas-"
```
nanux, line 404: Are these fixed constants? Any chance we can refer to them rather than magical string?

install.sh
```diff
@@ -393,6 +393,32 @@ set_current_context_k8s() {
   fi
 }
 
+scale_down() {
```
nanux, line 396: This is not the easiest read, would be useful to have some comments describing at least the flow.

nanux (approved):

nanux:
This wasn't happening before? Or it was just ignored?

PR #439, opened 2024-11-29
Fix for Terraform 1.10.0

bianchi2:
After Terraform has been upgraded to 10.1.0, our e2e [started failing](https://github.com/atlassian-labs/data-center-terraform/actions/runs/12077947907/job/33682321127#step:13:374). For some reason, Terraform started treating TV_VAR_* variables differently. 

So, instead of using `TF_VAR_ kinesis_log_producers_role_arns` (which is a json) env var, it's used in the tfvars file template that e2e test use.

Raised an issue upstream https://github.com/hashicorp/terraform/issues/36134

## Checklist


yzha645:
I saw [the fix](https://github.com/hashicorp/terraform/pull/36121) will be backported to 1.10, are we going to revert this change later?

yzha645 (approved):

PR #440, opened 2024-12-09
Adapt crowd-bitbucket e2e tests to Bitbucket 9.4.0

bianchi2:
* disable websudo
* use different api to get BITBUCKETSESSIONID (login call)

## Checklist

### DIFF
--- a/test/e2etest/crowd_test.go
+++ b/test/e2etest/crowd_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"net/http/cookiejar"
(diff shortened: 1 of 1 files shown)


test/e2etest/crowd_test.go
```diff
@@ -99,34 +100,44 @@ func addCrowdUserDirectory(t *testing.T, directoryName string, bitbucketURL stri
...
+	}
+	request.Header.Set("Content-Type", "application/json")
 	resp, err := client.Do(request)
 	if err != nil {
-		fmt.Println(err)
+		return "", fmt.Errorf("error making request: %v", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("authentication failed: status code %d", resp.StatusCode)
```
yzha645, line 131: Do you think it's worth passing out response body as well? We can then have more detailed message on the actual error.

test/e2etest/crowd_test.go
```diff
@@ -242,7 +253,7 @@ func crowdTests(t *testing.T, testConfig TestConfig, bitbucketURL string, crowdU
 
 	// get BITBUCKETSESSIONID to use in the header in subsequent calls
 	// even though basic auth works, atl_token is different each time
-	bitbucketSessionID := getBitbucketSessionID(bitbucketURL, "admin", testConfig.BitbucketPassword)
+	bitbucketSessionID, _ := getBitbucketSessionID(bitbucketURL, "admin", testConfig.BitbucketPassword)
```
yzha645, line 256: We're not processing err here even though we have various errors returning from the method?

yzha645 (approved):

